PR #7, opened 2021-05-25
Add version to packaging config

### LINKED ISSUE #2: Implement semantic versioning from tags
Semantic versioning with git tags: https://semver.org/

connortann:
This updates the build config to **version** the build according to the latest git tag.

Git tags are used as the single source of truth of the version. The version will then appear in several places:
- The built python package (i.e. when `python -m build` is run, the wheel file has the version in the name)
- The `resqpy.__version__` attribute
- The top right and bottom left of https://resqpy.readthedocs.io/

Closes #2 

References:

- https://packaging.python.org/guides/single-sourcing-package-version/
- https://github.com/pypa/setuptools_scm/

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -21,6 +21,9 @@ build
 dist
 *.egg-info
 
+# Dynamically-generated version
+resqpy/version.py
+
 # XSD files
 resqml/olio/data/Properties.xsd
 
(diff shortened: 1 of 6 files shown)


connortann:
@andy-beer  I believe the unit tests on this MR are failing due to something unrelated; a fix for that is in Merge Request #5.

PR #14, opened 2021-05-28
add_faults() bug fix for new fault intersecting existing fault

andy-beer:
fix of bug affecting split pillar data when adding a fault that intersects an existing fault


andy-beer:
I also added a line to .gitignore which probably only affects Mac users. (Should have been a separate branch and pull request, sorry!)

PR #18, opened 2021-05-28
Refactor DeviationSurvey class with class constructors

### LINKED ISSUE #27: Class constructor methods for creating objects
Presently, the init methods of many classes are quite complex, as they handle several things:

- Loading an existing object from disk 
- Creating a new object from input parameters
- Creating the object from e.g. other file formats

It could help to implement some separate `cls.from_file` class methods, to reduce the complexity of init methods

connortann:
An attempt to refactor an arbitrarily-chosen class `DeviationSurvey`, to explore simplifying the init method using class constructors e.g. `cls.from_dataframe()`.

Sets the `uuid` to be the only primary key, and `root_node` and `part` as properties that are dynamically evaludated using `uuid`.

Refactors the `load_xml` method considerably.

Also adds some more tests, and adds fixes to various other parts of the code to get the tests to pass.

Closes #27


connortann:
@emmanesbit FYI, I came across an issue with measured depths being in apparently the wrong place in the xml files for DeviationSurveys. Andy mentioned that you'd come across a similar issue recently?

emmanesbit:
> @emmanesbit FYI, I came across an issue with measured depths being in apparently the wrong place in the xml files for DeviationSurveys. Andy mentioned that you'd come across a similar issue recently?

Thanks Connor, your change fixes the issue I found in deviation survey. 

I ran into another issue related to deviation surveys in Trajectory, in load_from_xml, also in well.py. When defining ds_uuid (line 816 for me, though may have changed in your version) it was previously always returning None, i've updated the current line to `ds_uuid = bu.uuid_from_string(rqet.find_tag_text(rqet.find_tag(node, 'DeviationSurvey'), 'UUID'))` to fix this issue.

I'll merge this change once happy with testing, but as it might cause merge conflicts happy to update in your current branch if easier.

tests/test_well.py
```diff
@@ -61,3 +64,45 @@ def test_logs():
...
+   datum: resqpy.well.MdDatum
+   model, well_interp, datum, traj = example_model_with_well
+   epc_path = model.epc_file
+
+   # Create new survey
+   survey = resqpy.well.DeviationSurvey(
+      parent_model=model,
+      represented_interp=well_interp,
+      md_datum=datum,
+      measured_depths=[1,2,3],
```
andy-beer, line 84: These four numerical arrays would normally be floating point numbers – testing with integers is a curiosity but perhaps testing with float data would be better.

resqpy/well.py
```diff
@@ -317,222 +316,248 @@ class DeviationSurvey():
...
-         self.load_from_data_frame(data_frame, md_uom = length_uom, md_datum = md_datum)
-      elif deviation_survey_file:
-         self.load_from_ascii_file(deviation_survey_file,
-                                   space_separated_instead_of_csv = survey_file_space_separated,
-                                   md_uom = length_uom, md_datum = md_datum)
-      if self.uuid is None: self.uuid = bu.new_uuid()
 
+      # TODO: remove init from root_node, use just uuid
+      if root_node is not None:
+         warnings.warn("Argument root_node is deprecated, please use uuid")
```
andy-beer, line 373: Does this end up as a standard python logging warning message?

tests/test_well.py
connortann: I updated the write_hdf functino to ensure the MDs are cast as floats when saved to HDF5.

resqpy/well.py
connortann: It's a standard warning, not quite the same as logging.warning.

From the python [logging docs](https://docs.python.org/3/howto/logging.html), they suggest `warnings.warn`:
image

tests/test_well.py
andy-beer: The functino sounds like something discovered at CERN!

andy-beer (approved):

PR #28, opened 2021-06-01
Use autoclasstoc to separate out "high level" and "low level" methods

### LINKED ISSUE #16: Delineate most often used methods in the docs
For classes that have a large number of methods, it would be helpful to separate out the high-level and low-level methods somehow in the autosummary listing.

connortann:
Closes #16 

Uses the autoclasstoc library to enhance our Sphinx documentation with customised sections when listing out methods of a class: https://autoclasstoc.readthedocs.io/en/latest/index.html

Enables us to separate out high- and low-level methods in the docs, to make it a bit more tractable when the table gets very long. The enables us to promote a method to a special "High-level methods" section by including `:meta high-level:` at the end of the docstring.

The new-style documentation can be previewed here:
https://resqpy.readthedocs.io/en/split-docs/

### DIFF
--- a/docs/_templates/custom-class-template.rst
+++ b/docs/_templates/custom-class-template.rst
@@ -6,6 +6,6 @@
    :members:
    :show-inheritance:
    :inherited-members:
-   :special-members: __call__, __add__, __mul__
+   :special-members: __init__
 
    .. autoclasstoc::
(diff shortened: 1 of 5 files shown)


connortann:
@andy-beer here's a screenshot showing how the new docs look with a separate "high-level" section.

The name of the section(s) can be changed, this is just a first pass.

image

PR #32, opened 2021-06-02
Merging timeseries

jrt54:
New timeseries functions to merge together several timeseries into one sorted timeseries and return the resulting timeseries object. Also created a convenience routine to return datetime objects which may entail slicing off the "+Z" timezone character from the timestamps. 

This works only based off of timeseries uuids.

I created tests in tests/test_timeseries.py, but I'm not sure how to configure the CI workflow to run them automatically

### DIFF
--- a/resqpy/time_series.py
+++ b/resqpy/time_series.py
@@ -376,19 +376,21 @@ def time_series_from_list(timestamp_list, parent_model = None):
       time_series.add_timestamp(timestamp)
    return time_series
 
-def merge_timeseries_from_uuid(model, timeseries_uuid_iter, reverse=False):
-   """Create a TimeSeries object from an iteratable object of existing timeseries UUIDs of timeseries. iterable can be a list, array, or iteratable generator (model must be provided). reverse=True returns a timeseries object in ascending order. Reverse=None returns a timeseries without timestamps being sorted. Returns the new time series, the new time series uuid, and the list of timeseries objects used to generate the list"""
-    #assert(False)
+def merge_timeseries_from_uuid(model, timeseries_uuid_iter):
+   """Create a TimeSeries object from an iteratable object of existing timeseries UUIDs of timeseries. iterable can be a list, array, or iteratable generator (model must be provided). The new timeseries is sorted in ascending order. Returns the new time series, the new time series uuid, and the list of timeseries objects used to generate the list"""
+   
+   reverse=False 
+
    alltimestamps=set({})
    timeserieslist=[]
    for timeseries_uuid in timeseries_uuid_iter:
        timeseriesroot=model.root(uuid=timeseries_uuid)
        assert(rqet.node_type(timeseriesroot) == 'obj_TimeSeries')
        
-       #singlets=model.part_for_uuid(timeseries_uuid)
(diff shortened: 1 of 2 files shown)


connortann:
The static analysis failure is not your fault - looks like it wasn't set up correctly to run on forks.

I've updated the GitHub actions workflow in #35 ; so try rebasing your branch on top of master and see if it passes now.

connortann:
@andy-beer an issue is that python 3.6 does not support `datetime.fromisoformat` . It looks like we could get this working by installing this extra pip package:

https://github.com/movermeyer/backports.datetime_fromisoformat

However, it may be simpler to just drop support for python 3.6 . Two files need to be updated:
- setup.cfg , the classifiers section for PyPI
- github/workflows/ci-tests.yml , the "matrix strategy" section for running unit tests

resqpy/time_series.py
```diff
@@ -247,6 +247,18 @@ def extend_by_days(self, days):
...
+      datetimes=[]
+        
+      for timestamp in self.timestamps:
+         datetime_timestamp=timestamp
+         if timestamp.endswith('Z'): datetime_timestamp = timestamp[:-1]
+         datetimes.append(dt.datetime.fromisoformat(datetime_timestamp))
+        
+        
+
+      return datetimes
```
connortann, line 261: Perhaps a list comprehension is simpler here?

```suggestion
      return [dt.datetime.fromisoformat(t.rstrip('z')) for t in self.timestamps]
```

resqpy/time_series.py
```diff
@@ -364,6 +376,31 @@ def time_series_from_list(timestamp_list, parent_model = None):
...
+   for timeseries_uuid in timeseries_uuid_iter:
+       timeseriesroot=model.root(uuid=timeseries_uuid)
+       assert(rqet.node_type(timeseriesroot) == 'obj_TimeSeries')
+       
+       singlets=TimeSeries(model, time_series_root=timeseriesroot)
+       timeserieslist.append(singlets)
+       #alltimestamps.update( set(singlets.timestamps) )
+       alltimestamps.update( set(singlets.datetimes()) )
+   
+   if reverse is None:
```
connortann, line 395: Will reverse ever be None? It seems to be defined as False above

andy-beer:
I'm happy to drop support for python 3.6 and earlier, @connortann

connortann:
> I'm happy to drop support for python 3.6 and earlier, @connortann

OK, dropping support in #37 . @jrt54 , once that one is merged, please rebase this branch on top of master so the CI will run with the updated configuration.

resqpy/time_series.py
jrt54: No it won't. I originally tried to allow creation of a timeseries with an argument sorting or without sorting, but it seems like the returned timeseries generated from a list of a timestamps is almost always in ascending sorted order. So I removed the (reverse) argument.

jrt54:
Alright I rebased and pushed again, I believe that should have updated the PR

jrt54:
@connortann @andy-beer , is there any action waiting on me to go forward with this merge? All checks passed, but I see is a warning that I'm not authorized to merge.

PR #36, opened 2021-06-03
Optimising consolidation

andy-beer:
BOOTs Maps workflow is taking too long to consolidate large ensembles (time grows worse than linear with number of realisations). This change aims to provide some speedup when the calling code has some 'knowledge' about the ensemble.


andy-beer:
@emmanesbit I think these changes can be merged in. These optimisations are primarily for Patrick's mapping work for large ensembles.

emmanesbit (approved):
Looks good to me!

PR #42, opened 2021-06-03
WIP: Cross fault bug fix

andy-beer:
This change completely re-works the fault throw scaling function in the derived_model module. This function was not working property where faults intersected one another.

The creation of the column-pillar mapping in the Grid module has also been rewritten.


andy-beer:
Todo: add test functions

Issue #17, opened 2021-05-28
Consolidate unique keys of objects to use UUID


connortann:
An idea for a refactoring

Objects have a few different unique keys that are used throughout the code:

- UUID
- Part name (name of .xml file when unzipped)
- XML node

Would be good to standardise on using UUID

andy-beer:
The main change here is for resqpy classes to accept uuid as the primary key when instantiating from existing RESQML objects (ie. from xml). Use of the xml root node arg then triggers a deprecated warning. For backward compatibility, it is preferable for the new uuid arg to come after the existing root_node (or equivalent) arg.

PR #44, opened 2021-06-04
WIP: More extensive base class, handling XML/HDF attributes

### LINKED ISSUE #20: Base class for common attributes and methods
We could create a base class, and inherit from this in all other resqml class objects.

This could help reduce duplication of code, and standardise common attributes and methods.

The base class could handle things such as:
- uuid creation / lookup
- Full resqml type (e.g. WellboreFeatureInterpretation)
- `__eq__`, `__str__`, __repr__`, and other generic magic methods
- High-level methods for reading and saving to disk

It would also help make the classes more robust: we could have generic Model methods that would be compatible with any class that inherits from Base.

connortann:
An attempt to implement a Base Class for resqml objects, that handles loading and saving simple (not nested) properties to XML and HDF5

Builds on existing #68

Should help make progress towards #52 and #17

### DIFF
--- a/resqpy/olio/attributes.py
+++ b/resqpy/olio/attributes.py
@@ -1,0 +1,150 @@
+""" Easy attributes for defining what should be saved to XML & HDF5 """
+
+import logging
+from abc import abstractmethod
+from dataclasses import dataclass
+
+import resqpy.olio.xml_et as rqet
+import resqpy.olio.weights_and_measures as bwam
+from resqpy.olio.xml_namespaces import curly_namespace as ns
+
+
+logger = logging.getLogger(__name__)
+    
+
+class BaseAttribute:
+
+    xml_types = dict(
+        double           = dict(dtype=float, xml_ns='xsd'),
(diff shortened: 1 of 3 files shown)


connortann:
@andy-beer it would be great to get your thoughts on this initial implementation of the base class idea.

Some particular questions:

1. Do you agree with the choice of default attributes and methods, to be inherited by most simple resqml objects?
2. Are you happy with the api design that object is either created new or loaded from XML, depending on whether the `uuid` parameter is given? (An alternative might be to require "load_from_xml" to be called explicitly)
3. Does it make sense to you for `self.root` and `self.part` to be dynamically retrieved using `self.uuid`, or should they be treated some other way?
4. Is the order of operations correct for simple use cases, in particular the loading from XML and add_as_part?

Nb. It might be simplest review / merge #45 first, as this branch builds on tests written in that branch.

connortann:
@andy-beer  I've split these changes out into two separate chunks: the fundamental base class that I think is ready for your review is now in a separate PR , #68 .

connortann:
Closing for now, idea parked

PR #45, opened 2021-06-04
Add IO tests for resqpy.organise and resqpy.well objects

connortann:
Ready for review & merge

### DIFF
--- a/resqpy/organize.py
+++ b/resqpy/organize.py
@@ -12,6 +12,7 @@ log.debug('organize.py version ' + version)
 # from lxml import etree as et
 
 import math as maths
+import warnings
 
 import resqpy.olio.xml_et as rqet
 import resqpy.olio.uuid as bu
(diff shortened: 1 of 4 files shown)


resqpy/organize.py
```diff
@@ -74,23 +75,29 @@ def create_xml_has_occurred_during(model, parent_node, hod_pair, tag = 'HasOccur
 class OrganizationFeature():
    """Class for generic RESQML Organization Feature objects."""
 
-   def __init__(self, parent_model, root_node = None, extract_from_xml = True, feature_name = None, organization_kind = None):
+   def __init__(self, parent_model, uuid=None, root_node = None, extract_from_xml = True, feature_name = None, organization_kind = None):
```
andy-beer, line 78: Some resqpy code, and some Bifröst Labs code, might use a positional argument for the root_node when calling methods like this. Ideally we would scan these two repositories for such calls when making changes like this, otherwise something will break.

PR #49, opened 2021-06-07
Blocked well add interpretation and feature convenience

emmanesbit:
Added convenience function to quickly create wellbore interpretation and wellbore feature objects for a blocked well object. 
Other minor bug fixes.

### DIFF
--- a/resqpy/well.py
+++ b/resqpy/well.py
@@ -3013,6 +3013,10 @@ class BlockedWell:
       return xyz, rqet.uuid_for_part_root(self.trajectory.crs_root)
 
    def create_feature_and_interpretation(self):
+      """Instantiate new empty WellboreFeature and WellboreInterpretation objects
+      
+      Uses the Blocked well citation title as the well name
+      """
       self.wellbore_feature = rqo.WellboreFeature(parent_model=self.model,feature_name=self.trajectory.title,extract_from_xml=False)
       self.wellbore_interpretation = rqo.WellboreInterpretation(parent_model=self.model,extract_from_xml=False,wellbore_feature=self.wellbore_feature)
       self.trajectory.wellbore_interpretation = self.wellbore_interpretation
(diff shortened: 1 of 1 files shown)


resqpy/well.py
```diff
@@ -865,7 +866,7 @@ def load_from_xml(self, node, hdf5_source_model = None):
       md_datum_part = relatives_model.part_for_uuid(md_datum_uuid)
       assert md_datum_part, 'md datum part not found in model'
       self.md_datum = MdDatum(self.model, md_datum_root = relatives_model.root_for_part(md_datum_part))
-      ds_uuid = bu.uuid_from_string(rqet.find_tag(rqet.find_tag(node, 'DeviationSurvey'), 'UUID'))
+      ds_uuid = bu.uuid_from_string(rqet.find_tag_text(rqet.find_tag(node, 'DeviationSurvey'), 'UUID'))
```
connortann, line 869: I saw Andy wrote some functions for nested tag lookup, could perhaps be useful?

```suggestion
      ds_uuid = bu.uuid_from_string(rqet.find_nested_tags_text(node, ['DeviationSurvey', 'UUID']))
```

resqpy/well.py
```diff
@@ -3007,6 +3012,12 @@ def xyz_marker(self, active_only = True):
       xyz = self.trajectory.xyz_for_md(md)
       return xyz, rqet.uuid_for_part_root(self.trajectory.crs_root)
 
+   def create_feature_and_interpretation(self):
```
connortann, line 3015: Could have a docstring if it's a public method (or else, could give an underscore prefix to the name to make this a private method):

```suggestion
   def create_feature_and_interpretation(self):
      """Instantiate new empty WellboreFeature and WellboreInterpretation objects
      
      Uses the Blocked well citation title as the well name
      """
```

connortann (approved):
Added two comments for suggested tweaks, up to you whether to change. Otherwise looks good!

emmanesbit:
> Added two comments for suggested tweaks, up to you whether to change. Otherwise looks good!

I think I've accepted them both, good spot on the nested_tags_text and missing docstring, thanks!

PR #50, opened 2021-06-07
New convenience functions to add wellboreinterp and feature 

emmanesbit:
New convenience functions for wellboremarkerframe and trajectory, to allow easy adding of wellboreinterpretation and wellborefeature objects.

### DIFF
--- a/resqpy/lines.py
+++ b/resqpy/lines.py
@@ -15,6 +15,7 @@ from resqpy.olio.xml_namespaces import curly_namespace as ns
 import resqpy.olio.write_hdf5 as rwh5
 import resqpy.crs as rcrs
 import resqpy.olio.simple_lines as rsl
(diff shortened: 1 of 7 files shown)


emmanesbit:
@connortann - the changes to surface.py involve a lot of repetition, with exactly the same method appearing in TriangulatedSet, Mesh and PointSet. Keen to hear any suggestions for how to improve this (the best I could think of was adding a Parent class for these three classes to inherit from containing the shared method). Similarly the methods in well.py for BlockedWellbore, WellboreFrame and Trajectory are very similar, so it may be possible to improve these too

connortann (approved):
LGTM!

emmanesbit:
@connortann This should be ready to go now, apologies for the many commits to fix the unit tests.

PR #56, opened 2021-06-08
Fix docs version numbers, explicit docs dependencies

connortann:
The docs do not currently show the version number correctly when hyperlinks are copied and pasted around. I think this is because the `resqpy` package is not properly pip-installed when the docs are build - this should fix that issue.

Also suggest including the documentation dependencies explicitly in the package requirements (i.e. in `setup.cfg`), so that a developer can install everything needed with:

```
pip install .[tests,docs]
```

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -41,6 +41,7 @@ tests =
     pytest-cov
     flake8
 docs = 
+    sphinx
     autoclasstoc
 
 [tool:pytest]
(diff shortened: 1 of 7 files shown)


setup.cfg
```diff
@@ -40,6 +40,8 @@ tests =
     pytest
     pytest-cov
     flake8
+docs = 
+    autoclasstoc
```
connortann, line 44: This is the bit that enables `pip install -e .[tests,docs]` to work

andy-beer (approved):

PR #58, opened 2021-06-09
Point set write hdf5

andy-beer:
Small change to allow hdf5 file name and ext uuid to be defaulted in PointSet methods write_hdf5() and create_xml()


emmanesbit (approved):
Looks good

PR #60, opened 2021-06-10
addition of meta common indicators to docstrings, issue #34

andy-beer:
This change adds meta common lines to the docstrings for the more commonly used methods of large classes. This results in a separate grouping in readthedocs.


emmanesbit (approved):
Looks good to me

Issue #61, opened 2021-06-10
shared interpretation object for blocked well & trajectory required by RMS


andy-beer:
The current resqpy blocked well method for adding an interpretation and feature will share the feature with the trajectory object (if present) but create a separate interpretation object. RMS expects the interpretation object to be shared too. We can change the resqpy code to follow the RMS protocol.

PR #62, opened 2021-06-10
blocked well interpretation sharing trajectory interpretation if found; issue #61

andy-beer:
I haven't tested this yet Emma (@emmanesbit), but feel free to take a look anyway.


andy-beer:
Okay, the code change seems to be behaving as expected. I can update the block.epc example dataset (and perhaps s_bend too?).

andy-beer:
I've updated the block example dataset, but have decided to leave s_bend unchanged.

emmanesbit (approved):
Looks good to me

PR #68, opened 2021-06-14
Implement simple base class

### LINKED ISSUE #19: IPython representation methods for classes
We could add a few methods to aid use of resqpy in an interactive terminal or Jupyter notebook, such as:
- str
- repr
- repr_html

These could display information such as:
- RESQML class type (e.g. `WellboreFeatureRepresentation`)
- citation title
- uuid (?)
- relationships (?)

To avoid excessive duplication, these would presumably need to be inherited from a Base class, so this is probably dependent on #20

See https://ipython.readthedocs.io/en/stable/config/integrating.html

### LINKED ISSUE #20: Base class for common attributes and methods

connortann:
A second, simpler attempt to implement a simple Base Class for resqml objects, that handles the generic stuff like:

- Standardising on using `uuid` as the primary key
- Setting citation title and orginator attributes
- Pretty display methods, including HTML ipython

Closes #20
Closes #19

Similar to #44 , but with fewer changes for a first step.

In future, could be extended to implement loading and saving simple (not nested) properties to XML and HDF5

### DIFF
--- a/resqpy/olio/base.py
+++ b/resqpy/olio/base.py
@@ -11,7 +11,7 @@ import resqpy.olio.xml_et as rqet
 logger = logging.getLogger(__name__)
 
 
-class BaseResqml(metaclass=ABCMeta):
+class BaseResqpy(metaclass=ABCMeta):
     """Base class for generic RESQML objects
     
     Implements generic attributes such as uuid, root, part, title, originator.
@@ -21,7 +21,7 @@ class BaseResqml(metaclass=ABCMeta):
     
     Example use::
 
-        class AnotherResqmlObject(BaseResqml):
+        class AnotherResqmlObject(BaseResqpy):
             
             _content_type = 'obj_anotherresqmlobjectrepresentation'
 
--- a/resqpy/olio/base.py
+++ b/resqpy/olio/base.py
@@ -91,7 +91,7 @@ class BaseResqml(metaclass=ABCMeta):
         self.title = rqet.find_nested_tags_text(self.root, ['Citation', 'Title'])
         self.originator = rqet.find_nested_tags_text(self.root, ['Citation', 'Originator'])
 
-    def create_xml(self, title=None, originator=None, ext_uuid=None, add_as_part=True):
+    def create_xml(self, title=None, originator=None, add_as_part=True):
         """Write citation block to XML
         
         Args:
@@ -107,8 +107,6 @@ class BaseResqml(metaclass=ABCMeta):
         """
 
         assert self.uuid is not None
-
-        if ext_uuid is None: ext_uuid = self.model.h5_uuid()
 
         # Create the root node
         node = self.model.new_obj_node(self._content_type)
--- a/resqpy/olio/base.py
+++ b/resqpy/olio/base.py
@@ -91,7 +91,7 @@ class BaseResqpy(metaclass=ABCMeta):
         self.title = rqet.find_nested_tags_text(self.root, ['Citation', 'Title'])
         self.originator = rqet.find_nested_tags_text(self.root, ['Citation', 'Originator'])
 
-    def create_xml(self, title=None, originator=None, add_as_part=True):
+    def create_xml(self, title=None, originator=None, add_as_part=False):
         """Write citation block to XML
         
         Args:
@@ -112,16 +112,17 @@ class BaseResqpy(metaclass=ABCMeta):
         node = self.model.new_obj_node(self._content_type)
         node.attrib['uuid'] = str(self.uuid)
 
-        if add_as_part:
-            self.model.add_part(self._content_type, self.uuid, node)
-            assert self.root is not None   
-
         # Citation block
         if title: self.title = title
         if originator: self.originator = originator
         self.model.create_citation(
             root=node, title=self.title, originator=self.originator
         )
+        
+        if add_as_part:
+            self.model.add_part(self._content_type, self.uuid, node)
+            assert self.root is not None
+        
         return node
 
     # Generic magic methods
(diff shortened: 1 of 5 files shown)


resqpy/olio/base.py
```diff
@@ -0,0 +1,163 @@
...
+            add_as_part (boolean, default True): if True, the newly created xml node is added as a part
+                in the model
+
+        Returns:
+            node: the newly created root node
+        """
+
+        assert self.uuid is not None
+
+        if ext_uuid is None: ext_uuid = self.model.h5_uuid()
```
andy-beer, line 111: Not all classes require ext_uuid.

resqpy/olio/base.py
```diff
@@ -0,0 +1,163 @@
...
+    @root.setter
+    def root(self, value):
+        """Update self.uuid to match new root. Ensure root is added as a part"""
+
+        new_uuid = rqet.uuid_for_part_root(value)
+        if new_uuid is None:
+            raise ValueError("Cannot set uuid to be None")
+        self.uuid = new_uuid
+
+    def load_from_xml(self):
```
andy-beer, line 87: docstring could perhaps contain an explanation that this method only loads the common elements of the object, and that derived classes should complete the job.

resqpy/olio/base.py
```diff
@@ -0,0 +1,163 @@
...
+        self.model.create_citation(
+            root=node, title=self.title, originator=self.originator
+        )
+        return node
+
+    # Generic magic methods
+
+    def __eq__(self, other):
+        """Implements equals operator. By default, compare objects using uuid"""
+        return self.uuid is not None and self.uuid == getattr(other, "uuid", None)
```
andy-beer, line 131: resqpy.olio.uuid contains a function: matching_uuids() which is a safer way to compare uuids than ==

resqpy/olio/base.py
connortann: Good spot, that was a remnant of a previous bit of code where I was trying to handle writing HDF5 attributes. I'll remove this bit..

resqpy/olio/xml_et.py
```diff
@@ -124,6 +124,25 @@ def find_nested_tags(root, tag_list):
    return find_nested_tags(head, tag_list[1:])
 
 
+def find_nested_tags_cast(root, tag_list, dtype=None):
```
andy-beer, line 127: Let's move function to resqpy.olio.xml_et

resqpy/olio/base.py
```diff
@@ -0,0 +1,163 @@
...
+from abc import ABCMeta, abstractmethod
+
+import resqpy.olio.uuid as bu
+import resqpy.olio.xml_et as rqet
+
+
+logger = logging.getLogger(__name__)
+
+
+class BaseResqml(metaclass=ABCMeta):
```
andy-beer, line 14: I think we should call it BaseResqpy, rather than BaseResqml

resqpy/olio/base.py
connortann: Thanks, will do.

I was trying to understand the logic in some is_equivalent methods, maybe something more nuanced would be better?

```python
return isinstance(other, self.__class__) and any(
         self is other,
         bu.matching_uuids(self.uuid, other.uuid),
         all(
            self.title== getattr(other, "title", None),
            (equivalent_extra_metadata(self, other) or not check_extra_metadata))
      )
```

Or perhaps just comparing on uuids is simplest, given that is the point of them?

resqpy/olio/xml_et.py
connortann: It's already there

resqpy/olio/base.py
```diff
@@ -0,0 +1,163 @@
...
+        self.uuid = new_uuid
+
+    def load_from_xml(self):
+        """Load citation block from XML"""
+
+        # Citation block
+        self.title = rqet.find_nested_tags_text(self.root, ['Citation', 'Title'])
+        self.originator = rqet.find_nested_tags_text(self.root, ['Citation', 'Originator'])
+
+    def create_xml(self, title=None, originator=None, ext_uuid=None, add_as_part=True):
```
andy-beer: I think add_as_part should default to False here: it is preferable for the derived class method to finish populating the xml, and then add as part and create the relationships, in most cases.

resqpy/well.py
```diff
@@ -305,25 +305,9 @@ def __init__(self, parent_model, uuid=None, title=None, deviation_survey_root=No
...
-
-   @property
-   def part(self):
-      """Part corresponding to self.uuid"""
-      if self.uuid is None:
-         raise ValueError('Cannot get part if uuid is None')
-      return self.model.part_for_uuid(self.uuid)
+      super().__init__(
+         model=parent_model, uuid=uuid, title=title, originator=originator
+      )
```
andy-beer, line 311: Don't we want to call load_from_xml() if uuid (or deviation_survey_root) is provided?

resqpy/olio/base.py
connortann: I was hoping we could extend this class further ( in #44 ) so that in the majority of cases the base class would handle basic XML and HDF attributes, so that many simple derived classes would not need to implement anything to complete the job.

resqpy/well.py
connortann: Should be handled in the base class

PR #69, opened 2021-06-15
Faster implementation of self.part

connortann:
Switching to use `rqet.part_name_for_object` as suggested by Andy. Makes progress towards #17

### DIFF
--- a/tests/test_base.py
+++ b/tests/test_base.py
@@ -13,7 +13,8 @@ def test_base_creation(example_model):
     model, crs = example_model
     dummy = DummyObj(model=model, title=title)
 
-    # UUID and part should exist, but not root (yet)
+    # UUID should exist, and part (name) should be determinable,
+    # but root should not exist yet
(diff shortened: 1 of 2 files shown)


tests/test_base.py
```diff
@@ -13,15 +13,14 @@ def test_base_creation(example_model):
     model, crs = example_model
     dummy = DummyObj(model=model, title=title)
 
-    # UUID should exist, but not root or part
+    # UUID and part should exist, but not root (yet)
```
andy-beer, line 16: More strictly, "UUID should exist, and part (name) should be determinable, but root should not exist yet."

Most of the time we can use the term 'part' rather lazily to mean either the string that is a part name – whether or not that part has come into full existence and been added to a Model – or to mean a fully added part which has the part name as its key in the Model's internal dictionary. The base class part is providing the first of these.

andy-beer (approved):

PR #73, opened 2021-06-15
Simplify several FeatureInterpretation classes

connortann:
Adds inheritance for Geobody, Boundary, Frontier, GeologicUnit, FluidBoundary classes

### DIFF
--- a/resqpy/organize.py
+++ b/resqpy/organize.py
@@ -130,7 +130,7 @@ class OrganizationFeature(BaseResqpy):
 
       # Extra element for organization_kind
       if self.organization_kind not in ['earth model', 'fluid', 'stratigraphic', 'structural']:
-            raise ValueError(self.organization_kind)
+         raise ValueError(self.organization_kind)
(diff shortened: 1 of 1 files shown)


resqpy/organize.py
```diff
@@ -552,8 +404,6 @@ def create_xml(self, add_as_part = True, add_relationships = True, originator =
             self.model.create_reciprocal_relationship(rfuf, 'destinationObject', top_boundary_root, 'sourceObject')
             self.model.create_reciprocal_relationship(rfuf, 'destinationObject', base_boundary_root, 'sourceObject')
 
-      self.root_node = rfuf
```
andy-beer, line 555: Here and elsewhere: the root_node attribute is (from calling code perspective) replaced with root. This might break some things, though as we want to move away from calling code working with roots, maybe it's okay for things to break now!

andy-beer (approved):

resqpy/organize.py
connortann: root_node is still present as a deprecated alias

PR #76, opened 2021-06-15
Add high-level object iterator methods to Model

connortann:


resqpy/model.py
```diff
@@ -2913,14 +2913,15 @@ def wells(self):
       """
       import resqpy.organize  # Imported here to avoid circular imports
```
andy-beer, line 2914: for the versions of python we are using, circular imports should not cause a problem; however, it can still be good to have delayed imports so that the time is only gobbled up if the module is actually needed

resqpy/model.py
```diff
@@ -2951,6 +2954,20 @@ def md_datums(self):
...
+      """Iterable of all CRS objects associated with the model
+      
+      Yields:
+         crs: instance of :class:`resqpy.crs.CRS`
+
+      :meta common:
+      """
+      import resqpy.crs
+
+      uuids = self.uuids(obj_type = 'LocalDepth3dCrs')
```
andy-beer, line 2967: The resqpy Crs class also deals with the very similar RESQML class LocalTime3dCrs (if I remember correctly). I think this iterator should yield all the resqpy Crs objects possible, ie. after the depth ones, continue with the time ones.

resqpy/well.py
```diff
@@ -772,7 +774,7 @@ def load_from_xml(self, node, hdf5_source_model = None):
       assert md_datum_uuid is not None, 'failed to fetch uuid of md datum for trajectory'
       md_datum_part = relatives_model.part_for_uuid(md_datum_uuid)
       assert md_datum_part, 'md datum part not found in model'
-      self.md_datum = MdDatum(self.model, md_datum_root = relatives_model.root_for_part(md_datum_part))
+      self.md_datum = MdDatum(self.model, uuid=relatives_model.uuid_for_part(md_datum_part))
       ds_uuid = bu.uuid_from_string(rqet.find_nested_tags_text(node, ['DeviationSurvey', 'UUID']))
       if ds_uuid is not None:  # this will probably not work when relatives model is different from self.model
          ds_part = 'obj_DeviationSurveyRepresentation_' + str(ds_uuid) + '.xml'
```
andy-beer, line 780: could be improved by using the rqet function that builds the part name from content type and uuid

tests/conftest.py
```diff
@@ -20,36 +20,39 @@ def capture_logs(caplog):
...
 
 @pytest.fixture
-def example_model(tmp_path):
+def tmp_model(tmp_path):
+   """Example resqpy model in a temporary directory unique to each test"""
+
+   model = Model(create_basics=True)
+   model_path = str(tmp_path / 'tmp_model.epc')
+   model.set_epc_file_and_directory(model_path)
+   model.create_hdf5_ext(add_as_part=True, file_name=model_path[:-4] + '.h5')
```
andy-beer, line 29: there is now a convenience function in model.py: new_model(epc_path) which does lines 26, 28 & 29

tests/conftest.py
```diff
@@ -20,36 +20,39 @@ def capture_logs(caplog):
...
+   return model
+
+
+@pytest.fixture
+def example_model_and_crs(tmp_model):
    """ Returns a fresh RESQML Model and Crs, in a temporary directory """
```
andy-beer, line 35: Not sure this docstring is entirely accurate now that the tmp_model is being passed in

tests/test_s_bend.py
```diff
@@ -144,7 +144,7 @@ def df_trajectory(x, y, z):
 
    trajectory = rqw.Trajectory(model, md_datum = datum, data_frame = df, length_uom = 'm', well_name = 'ANGLED_WELL')
 
-   rqc.Crs(model, trajectory.crs_root).uuid == crs.uuid
+   rqc.Crs(model, uuid=trajectory.uuid).uuid == crs.uuid
```
andy-beer, line 147: This doesn't look right. The trajectory has its own uuid (trajectory.uuid) and it has a reference to a crs object. For historical reasons that is currently trajectory.crs_root (but needs at some point to be changed to trajectory.crs_uuid).  For now, I think we need something like:
   rqc.Crs(model, uuid = rqet.uuid_for_part_root(trajectory.crs_root)).uuid == crs.uuid

resqpy/well.py
```diff
@@ -220,6 +220,11 @@ def create_xml(self, add_as_part = True,
       self.root_node = datum
 
       return datum
+   
+   def __eq__(self, other):
```
andy-beer, line 224: Might be worth putting a comment in the code to indicate that this is a temporary implementation of __eq__ as a class like MdDatum is worthy of having a fully implemented version

Issue #74, opened 2021-06-15
Consolidation (reuse) at time of create_xml() call


andy-beer:
For simple resqpy classes (those with an is_equivalent() method), we can make life easier for calling code by having the option to consolidate (ie. reuse an existing part within the model) at the time of create_xml() being called. This has already been implemented for the Crs class, so this issue is to make similar changes to the other simple classes.  The change allows for code to call create_xml() for any simple object that is needed, without having to afterwards make a consolidation pass over the model.

Note though, that when such reuse is enabled, the create_xml() method may change the uuid of the resqpy object. Thus the create_xml() call should come immediately after instantiation of the resqpy object.

Issue #75, opened 2021-06-15
Consolidate & refactor resqpy.organize classes


connortann:
Refactor and add tests to classes in organize.py , to achieve:

- Common inherited basic methods (e.g. repr, ipython)
- Instantiating from UUID, with backward compatibility for root_node (supporting #17)
- Reduce duplication of code

Classes checklist:
- [x] OrganizationFeature
- [x] GeobodyFeature
- [x] BoundaryFeature
- [x] FrontierFeature
- [x] GeologicUnitFeature
- [x] FluidBoundaryFeature
- [x] RockFluidUnitFeature
- [x] TectonicBoundaryFeature
- [x] GeneticBoundaryFeature
- [x] WellboreFeature
- [x] FaultInterpretation
- [x] EarthModelInterpretation
- [x] HorizonInterpretation
- [x] GeobodyBoundaryInterpretation
- [x] GeobodyInterpretation
- [x] WellboreInterpretation

andy-beer:
Pull request #77 makes the changes for the remaining classes in organize.py, as well as introducing the reuse functionality.

PR #77, opened 2021-06-15
create_xml reuse argument for classes in organize.py; resolves issue #74

andy-beer:


andy-beer:
Note: this will be a breaking change with regard to the extract_from_xml argument which is being removed from the __init__ methods of all the classes in organize.py (ie. features and interpretations). The resqpy object will be loaded from xml if uuid is not None (or, deprecated, if root_node is not None). Otherwise, the resqpy object will be initialised from the other arguments.

resqpy/olio/base.py
```diff
@@ -90,22 +94,39 @@ def create_xml(self, title=None, originator=None, add_as_part=False):
...
+                is considered for reuse and if suitable the uuid of this object is modified
 
         Returns:
-            node: the newly created root node
+            node: the newly created root node, or reused root node as applicable
         """
 
         assert self.uuid is not None
 
+        if reuse:
```
connortann, line 117: Might it be simpler to have this "reuse" logic in a separate method  e.g. `self.try_reuse`, which would return `True` if a duplicate was found?

That way, the derived classes could do something like this in the overwritten `create_xml`:
```python
if reuse and self.try_reuse(): return self.root
ofn = super().create_xml(add_as_part=False, originator=originator)
```

Alternatively, perhaps the method could return a tuple of (node, was_reused), so the calling code could be:
```python
ofn, existing_found = super().create_xml(add_as_part=False, originator=originator, reuse=reuse)
if existing_found:
   return ofn
```

andy-beer: Yes, I had thought of the second alternative. The only thing is that then super().create_xml() will be returning something different from the derived classes create_xml(), which makes me think that we should give it a different name!

But looking at the code, I think your first suggestion is better. I'll go down that route and see how things shape up.

andy-beer:
This pull request also closes issue #75

resqpy/olio/base.py
```diff
@@ -99,13 +128,13 @@ def create_xml(self, title=None, originator=None, add_as_part=False):
                 in the model
 
         Returns:
-            node: the newly created root node
+            node: the newly created root node, or reused root node as applicable
```
connortann, line 131: I think this update no longer applies

connortann:
This looks great. I think we should try to add some simple IO tests into test_organize.py whilst we're at it for the remaining classes, to ensure they instantiate and load/save attributes as expected

resqpy/organize.py
```diff
@@ -408,148 +365,85 @@ def create_xml(self, add_as_part = True, add_relationships = True, originator =
...
+   feature_name = _alias_for_attribute("title")
+   valid_kinds = ('horizon', 'geobody boundary')
 
-      self.model = parent_model
-      self.root_node = root_node
-      self.uuid = None
+   def __init__(self, parent_model, root_node = None, uuid = None, kind = None, feature_name = None):
+      """Initialises a genetic boundary feature (horizon or geobody boundary) organisational object."""
       self.kind = kind
       self.absolute_age = None   # (timestamp, year offset) pair, or None
```
connortann, line 421: Should the `age` attribute be settable from the init method?

resqpy/organize.py
```diff
@@ -901,23 +719,18 @@ def is_equivalent(self, other, check_extra_metadata = True):
...
       # note: related organization feature node should be created first and referenced here
 
-      emi = self.model.new_obj_node('EarthModelInterpretation')
+      if not self.title: self.title = self.organization_feature.feature_name
+      if title_suffix: self.title += ' ' + title_suffix
+
+      if reuse and self.try_reuse(): return self.root
+      emi = super().create_xml(add_as_part=False, originator=originator)
 
       if self.organization_feature is not None:
```
connortann, line 735: What happens if organization_features is None : should a ValueError be raised?

resqpy/organize.py
```diff
@@ -955,69 +756,52 @@ def create_xml(self, organization_feature_root = None,
...
-         self.genetic_boundary_feature = genetic_boundary_feature
-         self.domain = domain
-         self.boundary_relation_list = boundary_relation_list
-         if self.boundary_relation_list is not None:
-            for boundary_relation in self.boundary_relation_list:
-               assert boundary_relation in ['conformable', 'unconformable below and above',
-                                            'unconformable above', 'unconformable below']
-
-      if self.uuid is None: self.uuid = bu.new_uuid()
+      self.boundary_relation_list = boundary_relation_list.copy()
```
connortann, line 784: This will throw an exception when the default args are left: `None` cannot be copied

resqpy/organize.py
andy-beer: I believe this is a geological age, though a little strange that it is stored as a timestamp, year offset pair – that is probably RESQML being very rigorous about how it handles a point in time!

andy-beer: And yes, I guess we should support it, even though no dataset has yet made use of it.

resqpy/organize.py
andy-beer: Ah, that is a bug then

resqpy/organize.py
andy-beer: I suggest we note this as a todo and leave it as is for now.

resqpy/organize.py
andy-beer: I can fix this one.

resqpy/olio/base.py
andy-beer: correct, I'll make the reversion

tests/test_organize.py
```diff
@@ -82,3 +94,62 @@ def test_RockFluidUnitFeature(tmp_model):
...
+
+    gen = rqo.GeneticBoundaryFeature(tmp_model, kind='horizon')
+    hor = rqo.HorizonInterpretation(
+        tmp_model, genetic_boundary_feature=gen, sequence_stratigraphy_surface='maximum flooding'
+    )
+
+    gen.create_xml()
+    hor.create_xml()
+
+    hor2 = rqo.HorizonInterpretation(tmp_model, uuid=gen.uuid)
```
andy-beer, line 142: the uuid arg should be hor.uuid I think

PR #85, opened 2021-06-17
Updates to prefer uuids rather than parts or roots

connortann:
Supports #17

### DIFF
--- a/resqpy/model.py
+++ b/resqpy/model.py
@@ -2912,13 +2912,10 @@ class Model():
       """
       import resqpy.organize  # Imported here for speed, module is not always needed
 
-      # TODO: instantiate using UUIDs
-      parts = self.parts_list_of_type('WellboreInterpretation')
-      for part in parts:
-         well_root = self.root_for_part(part)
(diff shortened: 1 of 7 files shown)


resqpy/model.py
```diff
@@ -2912,13 +2912,10 @@ def iter_wellbore_interpretations(self):
...
       import resqpy.organize  # Imported here for speed, module is not always needed
 
-      # TODO: instantiate using UUIDs
-      parts = self.parts_list_of_type('WellboreInterpretation')
-      for part in parts:
-         well_root = self.root_for_part(part)
-         well = resqpy.organize.WellboreInterpretation(self, root_node=well_root)
-         yield well
-
+      uuids = self.uuids(obj_type='WellboreInterpretation')
```
andy-beer, line 2915: I believe that Model.uuids(), parts(), roots() and titles() should now be guaranteed to return a list (possibly empty), so it should be safe to remove the `if uuids:` here.

andy-beer (approved):

PR #87, opened 2021-06-18
base and reuse workover of StringLookup, PropertyKind, TimeSeries & Crs classes

### LINKED ISSUE #79: StringLookup, TimeSeries and Crs to use BaseResqpy and reuse functionality
This issue is to modernise the resqpy StringLookup and TimeSeries classes to inherit from BaseResqpy; and also to add the reuse argument to their create_xml() methods.

The Crs class already has reuse; however, it is to be modified to inherit from BaseResqpy

andy-beer:
The Crs class has already had some reuse functionality updated.

Closes #79

### DIFF
--- a/resqpy/crs.py
+++ b/resqpy/crs.py
@@ -105,15 +105,16 @@ class Crs(BaseResqpy):
       root_node = self.root
       flavour = rqet.node_type(root_node)
       assert flavour in ['obj_LocalDepth3dCrs', 'obj_LocalTime3dCrs']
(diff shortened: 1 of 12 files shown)


connortann:
Looks good. May I suggest adding some simple unit tests for `StringLookup`, `PropertyKind` and `olio.consolidation`?

andy-beer:
I've added a test for StringLookup and another for TimeSeries; PropertyKind already has a simple test which includes some reuse testing; similarly for Crs.

connortann (approved):

PR #89, opened 2021-06-20
Initial code for Property class

andy-beer:
This change introduces a Property class which sits alongside PropertyCollection but behaves more like other resqpy classes. It aims to be easier to use than PropertyCollection when calling code is only interested in one property at a time.

### DIFF
--- a/resqpy/property.py
+++ b/resqpy/property.py
@@ -3066,7 +3066,6 @@ class Property(BaseResqpy):
          new resqpy Property object
       """
 
-      self.part = None
       self.collection = PropertyCollection()
       super().__init__(model = parent_model, uuid = uuid, title = title, originator = originator)
       if support_uuid is not None:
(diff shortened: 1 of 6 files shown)


resqpy/property.py
```diff
@@ -3039,6 +3039,332 @@ def basic_static_property_uuids_dict(self, realization = None, share_perm_parts
...
+      return self.collection.constant_value_for_part(self.part)
+
+   def from_array(self, cached_array, source_info, keyword,
+                  support_uuid, property_kind = None, local_property_kind_uuid = None,
+                  indexable_element = None, facet_type = None, facet = None,
+                  discrete = False, uom = None, null_value = None,
+                  time_series_uuid = None, time_index = None,
+                  realization = None, count = 1, const_value = None,
+                  string_lookup_uuid = None, find_local_property_kind = True, extra_metadata = {}):
+      """Populates this (empty) property from a numpy array and associated metadata.
```
connortann, line 3242: For consistency, this may make sense as a class constructor method, like from_singleton.
May be worth having this function located in the code next to the other constructor methods, for readability..

resqpy/property.py
```diff
@@ -3039,6 +3039,332 @@ def basic_static_property_uuids_dict(self, realization = None, share_perm_parts
...
+      )
+
+      # Edit properties of parent collection
+      prop.collection.has_single_property_kind_flag = True
+      prop.collection.has_single_indexable_element_flag = True
+      prop.collection.has_multiple_realizations_flag = False
+
+      return prop
+
+   def array_ref(self, dtype = None, masked = False, exclude_null = False):
```
connortann, line 3121: A tentative suggestion here: to be more "pythonic" and fit in with the look-and-feel of other scientific computing libraries, this method could be named `values()` .

If I didn't read the docstring, I might guess that `array_ref` was some kind of reference to the HDF5 location; whereas it actually seems to be the most important method that returns the actual property data

andy-beer: I'd rather postpone switching to "values" as we would have to change several method names throughout resqpy (and go through a deprecation cycle for the array_ref or similar names). I'll create a separate issue for now.

connortann (approved):

PR #94, opened 2021-06-22
Well log creation and conversion

connortann:
Adds a method `WellLogCollection.add_log`, to enable well logs to be created from scratch and saved to disk.

Supporting changes:
- Unit tests, including for previous functionality for export of logs as pandas or lasio.LasFile
- rename of `log.name` to `log.title`
- rename of `WellLogCollection.logs()` to `WellLogCollection.iter_logs()`
- Update of a few other tests, to stop using deprecated keywords (and thus reduce number of pytest warnings).

### DIFF
--- a/resqpy/property.py
+++ b/resqpy/property.py
@@ -792,6 +792,15 @@ class PropertyCollection():
 
       return list(self.dict.keys())
 
+   def uuids(self):
+      """Return list of uuids in this collection.
+
+      returns:
+         list of uuids being the members of this collection; there is one uuid per property array
+
+      :meta common:
+      """
+      return [self.model.uuid_for_part(p) for p in self.dict.keys()]
 
    def selective_parts_list(self,
                             realization = None,
@@ -3927,7 +3936,7 @@ class WellLogCollection(PropertyCollection):
             print(log.title)
       """
       
-      return (WellLog(collection=self, part=part) for part in self.parts())
+      return (WellLog(collection=self, uuid=uuid) for uuid in self.uuids())
 
    def to_df(self, include_units=False):
       """ Return pandas dataframe of log data
@@ -4010,19 +4019,20 @@ class WellLogCollection(PropertyCollection):
 class WellLog:
    """ Thin wrapper class around RESQML properties for well logs """
 
-   def __init__(self, collection, part):
+   def __init__(self, collection, uuid):
       """ Create a well log from a part name """
 
       self.collection: PropertyCollection = collection
       self.model = collection.model
-      self.part = part
+      self.uuid = uuid
 
+      part = self.model.part_for_uuid(uuid)
       indexable = self.collection.indexable_for_part(part)
       if indexable != 'nodes':
          raise NotImplementedError('well frame related property does not have nodes as indexable element')
 
       #: Name of log
-      self.title = self.model.citation_title_for_part(part)  # make this title
+      self.title = self.model.citation_title_for_part(part)
 
       #: Unit of measure
       self.uom = self.collection.uom_for_part(part)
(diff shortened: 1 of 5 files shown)


resqpy/property.py
```diff
@@ -792,6 +792,15 @@ def parts(self):
...
 
+   def uuids(self):
+      """Return list of uuids in this collection.
+
+      returns:
+         list of uuids being the members of this collection; there is one uuid per property array
+
+      :meta common:
+      """
+      return [self.model.uuid_for_part(p) for p in self.dict.keys()]
```
andy-beer, line 803: Could alternatively make use of the uuid stored in the dictionary entry:
[self.uuid_for_part(p) for p in self.dict.keys()]
or
[values[2] for values in self.dict.values()]

resqpy/property.py
```diff
@@ -3863,21 +3872,71 @@ def __init__(self, frame=None, property_set_root=None, realization=None):
...
+      """Add a well log to the collection, and optionally save to HDF / XML
+      
+      Args:
+         title (str): Name of log, typically the mnemonic
+         data (array-like): log data to write. Must have same length as frame MDs
+         unit (str): Unit of measure
+         discrete (bool): by default False, i.e. continuous
+         case_sensitive_units (bool): If True, consider case when parsing units.
+         realization (int): If given, assign data to a realisation.
+
```
andy-beer, line 3889: docstring missing line for write arg

resqpy/property.py
```diff
@@ -3863,21 +3872,71 @@ def __init__(self, frame=None, property_set_root=None, realization=None):
...
+      Returns:
+         uuids: list of uuids of newly added properties. Only returned if write=True.
```
andy-beer, line 3891: As the method adds exactly one property, might be better to extract the single uuid from the list and return the uuid

resqpy/well.py
```diff
@@ -1412,16 +1412,20 @@ def __init__(self, parent_model, frame_root=None, trajectory=None, mds=None, rep
 
       #: All logs associated with the wellbore frame; an instance of :class:`resqpy.property.WellLogCollection`
       self.logs = None
-
+
```
andy-beer, line 1415: I will be stripping out trailing whitespace at some point!

resqpy/well.py
```diff
@@ -3610,42 +3611,17 @@ def add_las_to_trajectory(las: lasio.LASFile, trajectory, realization=None,
...
+         case_sensitive_units=case_sensitive_units,
+         realization=realization
       )
-
-   # Step 3. write imported list arrays to hdf5 file
-   collection.write_hdf5_for_imported_list()
-   # Step 4. create resqml xml nodes for imported list arrays and add as parts to model
-   # And step 5. include newly added parts in collection
-   collection.create_xml_for_imported_list_and_add_parts_to_model()
+      # TODO: for speed, only write on last log
```
andy-beer, line 3624: I think for the speed improvement, it is just a case of setting write=False in the call to add_log, and then reverting the code removal just above here.

andy-beer (approved):
I've made a few comments but it's basically fine.

resqpy/property.py
connortann: I don't follow your comment - isn't this already being done that way?

resqpy/property.py
connortann: I think it may return more than one uuid:

```python
collection.add_log(... , write=True)   # Returns 1 uuid

collection.add_log(... , write=False)  # Returns None
collection.add_log(... , write=False)  # Returns None
collection.add_log(... , write=True)   # Returns 3 uuids

```

resqpy/well.py
connortann: That's right - I'll add a comment to the docstring.

PR #97, opened 2021-06-23
Add method to export model as a graph

### LINKED ISSUE #22: Graph visualisation of "what is in a model"
Could potentially be done with networkx, graphviz and/or holoviews. 

May lead to adding a significant dependency to resqpy, so could perhaps be optional.

connortann:
Closes #22

Example of resulting graph, plotted with networkx and holoviews:

image

### DIFF
--- a/resqpy/model.py
+++ b/resqpy/model.py
@@ -3052,13 +3052,19 @@ class Model():
       nodes = {}
       edges = set()
 
-      for uuid in map(str, self.uuids()):
+      if uuids_subset is None:
+         uuids_subset = self.uuids()
+      
+      uuids_subset = set(map(str, uuids_subset))
+
+      for uuid in uuids_subset:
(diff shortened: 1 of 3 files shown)


resqpy/model.py
```diff
@@ -3006,6 +3006,68 @@ def sort_parts_list_by_timestamp(self, parts_list):
...
+         - nodes: dict mapping uuid to attributes (e.g. citation title)
+         - edges: set of unordered pairs of uuids, representing relationships
+
+      :meta common:
+      """
+      nodes = {}
+      edges = set()
+
+      if uuids_subset is None:
+         uuids_subset = self.uuids()
```
connortann, line 3056: It would be nice if `Model.uuids()` supported a `uuid_list` argument, similar to the `parts_list` of `Model.parts()`, that had robust handling of uuid comparisons

PR #99, opened 2021-06-24
Create codeql-analysis.yml

connortann:
Creates alerts, visible here:

https://github.com/bp/resqpy/security/code-scanning

### DIFF
--- a/.github/workflows/codeql-analysis.yml
+++ b/.github/workflows/codeql-analysis.yml
@@ -1,0 +1,71 @@
+# For most projects, this workflow file will not need changing; you simply need
+# to commit it to your repository.
+#
+# You may wish to alter this file to override the set of languages analyzed,
+# or to provide custom queries or build logic.
+#
+# ******** NOTE ********
+# We have attempted to detect the languages in your repository. Please check
+# the `language` matrix defined below to confirm you have the correct set of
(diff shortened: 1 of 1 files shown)


connortann:
Nb. The "1 analysis not found" message is expected the first time this analysis runs. The step expects to compare the code scanning results with a previous run in `master`, but when doing first-time setup there is obviously no existing analysis to compare against.

andy-beer (approved):
Let's see how slow it is with normal merges.  If it's going to take minutes, is there any simple way we could make the code vulnerability analysis less frequent ? Either a separate manual button push once in a while, or when we make a new tag perhaps?

connortann:
Yes absolutely. That is controlled by the top bit of the YAML file, it is easy to configure. I have gone with the suggested defaults:

```yaml
on:
  push:
    branches: [ master ]     # Triggered after a pull request is merged into master
  pull_request:
    branches: [ master ]     # Triggered when a pull requests is opened or updated
  schedule:
    # 10:16 on Fridays       # Also runs on a schedule, to pick up new alert rules or 3rd party issues
    - cron: '16 10 * * 5'
```

There is definitely some advantage in having code scanning before a PR is merged, to catch issues early. But as a minimum, we could have it scanned before making a release and using it.

PR #102, opened 2021-06-24
Implement unit conversion for a wider range of units

connortann:
Adds a pint unit registry, that can be used to convert between a wide range of units.

Eventually, should be able to handle a large set of units, including all valid resqml uoms.

### DIFF
--- a/resqpy/olio/data/unit_registry.txt
+++ b/resqpy/olio/data/unit_registry.txt
@@ -51,6 +51,7 @@ month = year / 12
 # Volume
 cm3 = cm**3
 ft3 = foot**3
+m3 = m**3
 poreunits = cm3 / (cm3 * 100) = pu
 percent = 0.01 = %
 bbl = (m**3) * 0.1234
--- a/resqpy/olio/data/unit_registry.txt
+++ b/resqpy/olio/data/unit_registry.txt
@@ -51,7 +51,8 @@ month = year / 12
 # Volume
 cm3 = cm**3
 ft3 = foot**3
-pu = cm3 / (cm3 * 100)  # pore units
+poreunits = cm3 / (cm3 * 100) = pu
+percent = 0.01 = %
 bbl = (m**3) * 0.1234
 
 # Force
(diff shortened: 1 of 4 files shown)


resqpy/olio/data/unit_registry.txt
```diff
@@ -0,0 +1,98 @@
...
+    square_yard = yard ** 2 = sq_yd
+    square_mile = mile ** 2 = sq_mi
+
+    cubic_inch = in ** 3 = cu_in
+    cubic_foot = ft ** 3 = cu_ft = cubic_feet
+    cubic_yard = yd ** 3 = cu_yd
+@end
+
+@group USCSLengthSurvey
+    survey_foot = 1200 / 3937 * meter = sft = ft[US]
```
andy-beer, line 97: Have we checked that a 'survey foot' is actually the same as 'ft[US]'?

tests/test_units.py
```diff
@@ -0,0 +1,65 @@
...
+        bwam.parse_unit('foobar')
+
+
+def test_strip_scaling_prefix():
+    assert bwam.strip_scaling_prefix("ft") == (1, "ft")
+    assert bwam.strip_scaling_prefix("100 ft") == (100, "ft")
+    assert bwam.strip_scaling_prefix("1/32 ft") == (1/32, "ft")
+    assert bwam.strip_scaling_prefix("1e9 ft") == (1e9, "ft")
+    assert bwam.strip_scaling_prefix("1.2 ft") == (1.2, "ft")
+
```
andy-beer, line 45: perhaps add tests for prefixes of form: '0.001' and '1E6'

andy-beer (commented):
Let's talk it over again – I still have mixed feelings!

connortann:
Thanks for the discussion Andy! I'll close this PR for now without merging. To recap our current plan:

1. Implementing RESQML units of measure explicitly from a built JSON database looks more promising than using pint
2. We could repurpose some of this code in other repositories that would benefit from broader, general-purpose unit conversion
3. We can cherry-pick the tests from this branch and use them with our future alternate implementation

PR #103, opened 2021-06-24
Surface upheaval

andy-beer:
Bringing resqpy surface classes in line with BaseResqpy and uuid way of working.  This will be a breaking change, removing the extraneous extract_from_xml arguments and also unused parent node argument to create_xml() methods.


connortann (approved):
LGTM!

PR #105, opened 2021-06-25
Changes to fault.py (GridConnectionSet) to use BaseResqpy and uuid; closes #104

### LINKED ISSUE #104: fault upheaval
The GridConnectionSet class needs to be modified to inherit from BaseResqpy and switch to using uuid for initialisation from xml.

andy-beer:
This change brings the resqpy GridConnectionSet class in line with others inheriting from BaseResqpy

### DIFF
--- a/resqpy/model.py
+++ b/resqpy/model.py
@@ -1535,8 +1535,8 @@ class Model():
       gcs_list = []
       gcs_parts_list = self.parts_list_of_type('GridConnectionSetRepresentation')
       for part in gcs_parts_list:
-         gcs_root = self.root_for_part(part)
-         gcs = rqf.GridConnectionSet(self, connection_set_root = gcs_root)
+         gcs_uuid = self.uuid_for_part(part)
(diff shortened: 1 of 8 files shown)


resqpy/model.py
```diff
@@ -1535,8 +1535,8 @@ def grid_connection_set_list(self):
       gcs_list = []
```
connortann, line 1535: For consistency should this be one of the "iter" methods, i.e. `model.iter_grid_connection_sets` ?

connortann:
Looks good. May I suggest taking this opportunity to add some unit tests for the bits in this diff that aren't covered? I think:

- The `find_faces_to_respresent_surface_xxx()` functions
- `fault.add_connection_set_and_tmults`

Or if adding those tests that would make more sense in a separate PR, then LGTM.

resqpy/model.py
andy-beer: Yes, good call, seeing as we have other breaking changes that will need attention in calling code, I'll change this to a generator.

PR #112, opened 2021-06-29
Implement full set of RESQML units and quantities

### LINKED ISSUE #95: Support full set of RESQML units of measure
Presently, common units such as "v/v" are lost and stored as "EUC".

A really exciting option could be to include pint, and include a short config file to handle the most common units used in reservoir modelling. This could be (hopefully) an elegant way to allow a wide range of common units, with simple configuration - and could be a "selling point" of resqpy more broadly.

https://pint.readthedocs.io/en/stable/

connortann:
Supports #95

Adds a database of the full resqml unit system, in the form of `olio/data/properties.json`, and a corresponding build script.

Implements a few simple functions that read this:
- `bwam.valid_uoms()`  -> returns all valid resqml uoms
- `bwam.valid_property_kinds()` -> returns all valid resqml property kinds

Adds logic for unit conversion.


resqpy/olio/weights_and_measures.py
```diff
@@ -58,27 +75,59 @@ def rq_uom(units):
...
+   # Backwards formula: x=(A-Cy)/(Dy-B)
+   # All current units have D==0
 
-   # NB: other length units are supported by resqml
-   if units.lower() in ['m', 'metre', 'metres']: return 'metres'
-   if units.lower() in ['ft', 'foot', 'feet', 'ft[us]']: return 'feet'
-   assert(False)  # unrecognised length units
+   A1, B1, C1, D1 = get_conversion_factors(rq_uom(unit_from))
+   A2, B2, C2, D2 = get_conversion_factors(rq_uom(unit_to))
+   y = (A1 + (B1*x)) / (C1 + (D1*x))
```
Connossor, line 102: Something missing: this function should check that the units actually share the same base units, else the conversion will be nonsense.

The function 'get_conversion_factors' could perhaps also return the base unit.

[This is Connor on mobile]

andy-beer: Yes, agreed it would be good to do a check here – and bundling the base units into the results of get_conversion_factors() seems appropriate. Note that this convert function needs to be fast.

I would also add a line along the lines of _if unit_from == unit_to: return x_

resqpy/olio/weights_and_measures.py
```diff
@@ -58,27 +75,59 @@ def rq_uom(units):
...
-   rq_list = []
-   for u in units_list: rq_list.append(rq_uom(u))
-   return rq_list
+   Returns:
+      Converted value(s)
 
+   """
+   # TODO: robust handling of errors. At present, bad units are treated as "EUC", and will fail silently.
 
-def p_length_unit(units):
```
andy-beer, line 75: The p_*_units refer to pagoda units – a historic unit handling that got bundled into this module.  It doesn't have anything to do with RESQML, so shouldn't live in resqpy. However, pagoda will fail to run without changes if we strip the functions out of here. I suggest when we put this live we either keep a copy of the old file and move it into the pagoda source code area, or write a wrapper there.

andy-beer: A more ambitious project would be to re-write pagoda to be properly based on resqpy – a much bigger undertaking!

resqpy/property.py
```diff
@@ -4878,22 +4878,18 @@ def validate_uom_from_string(input_unit, case_sensitive=True):
    if not input_unit:
       return default
 
-   # Get set of all valid uoms
-   valid_uoms = set(bwam.properties_data()['uoms'])
-
    if case_sensitive:
-      if input_unit in valid_uoms:
+      if input_unit in bwam.valid_uoms():
```
andy-beer, line 4882: Perhaps we should emit a warning if a valid uom is not found?

resqpy/olio/weights_and_measures.py
```diff
@@ -275,16 +310,52 @@ def convert_flow_rates(a, from_units, to_units):
...
+      ValueError if either uom is not a valid resqml uom
+   """
+   if uom not in valid_uoms():
+      raise ValueError(f"{uom} is not a valid uom")
+   uoms_data = _properties_data()["units"][uom]
+   try:
+      a, b, c, d = uoms_data["A"], uoms_data["B"], uoms_data["C"], uoms_data["D"]
+   except KeyError:  # Base units do not have factors defined
+      a, b, c, d = 0, 1, 1, 0
+   return a, b, c, d
```
andy-beer, line 332: agreed that this function should also return the base uom

tests/test_property.py
```diff
@@ -80,22 +105,52 @@ def test_create_Property_from_singleton_collection(tmp_model):
...
+   # Duck typing should work
+   value = np.array([1,2,3])
+   expected = np.array([1000, 2000, 3000])
+   result = bwam.convert(value, unit_from="km", unit_to="m")
+   assert_array_almost_equal(result, expected)
+
+
+def test_conversion_factors_are_numeric():
+   for uom in bwam.valid_uoms():
+      factors = bwam.get_conversion_factors(uom)
```
andy-beer, line 142: If base uom is also returned, we could check here that it is a valid uom.

andy-beer (approved):
Looks good – I've posted a few minor comments

resqpy/property.py
connortann: Could we go further and raise an Exception if a valid unit is not found? I'm a bit worried about "failing silently".

For example, suppose a property is stored in e.g. "pore units", which is not a valid uom. This might get saved as "EuC", which can then be converted into units of "%". This will then be wrong by a factor of 100.

On a related note - I think there is some duplication between `property.validate_uom_from_string` and `bwam.rq_uom`. Do you think they should be consolidated?

PR #126, opened 2021-07-05
Consolidate unit conversion code, add quantities

connortann:
Expands the list of ALIASES to cover all previously written logic

Adds a "quantity" keywork to `rq_uom` and `convert` functions, to ensure uom is consistent with desired quantity

Adds RESQML-compliant in-place conversion (but only for trivial conversions that are a scaling factor)

Replaces old conversion functions with conversions from the RESQML standard

Renames `test_wam` into `test_weights_measures`, and moves some relevant unit tests there.

"Unit tests"... pun fully intended


andy-beer (approved):
Looks good

connortann:
@andy-beer  I've pushed more changes, so you may wish to re-review

andy-beer (approved):
It's great to see all those specific convert functions simplified to a call to the general conversion function!

andy-beer:
Yes, good to merge from my side.

PR #127, opened 2021-07-05
Adds type hints for arguments in crs module

### LINKED ISSUE #113: Add type annotations and type checking
Based on the setup.cfg this package is python3.7+, so there is no reason not to add type annotations.

andy-beer:
Closes #113


connortann:
Following discussion in #113 - I'm tempted to suggest parking these changes for now, unless we decide to full go for rolling out type hints across entire lib

andy-beer:
Agreed to park these changes.  Closing this pull request without merging.

resqpy/crs.py
```diff
@@ -30,12 +35,23 @@ def resqml_type(self):
...
-                x_offset = 0.0, y_offset = 0.0, z_offset = 0.0,
-                rotation = 0.0, xy_units = 'm', z_units = 'm',
-                z_inc_down = True, axis_order = 'easting northing',
-                time_units = None, epsg_code = None,
-                title = None, originator = None, extra_metadata = None):
+   def __init__(self,
+                parent_model: 'rq.Model',
+                crs_root = None,  # deprecated
+                uuid: Optional[uuid.UUID] = None,
+                x_offset: Optional[float] = 0.0,
```
connortann, line 42: Is this really optional? The default is `0.0` ; what would `None` mean?

resqpy/crs.py
```diff
@@ -30,12 +35,23 @@ def resqml_type(self):
...
+                y_offset: Optional[float] = 0.0,
+                z_offset: Optional[float] = 0.0,
+                rotation: Optional[float] = 0.0,
+                xy_units: Optional[str] = 'm',
```
connortann, line 46: Similarly, this looks like an argument which is probably not optional?

resqpy/crs.py
andy-beer: Ah, I assumed that Optional means that calling code doesn't need to pass a value, ie. it is a keyword argument with a default value.  Almost all the Crs attributes are required by RESQML. The exceptions, from memory, are time_units (only present when z units are time), epsg code, parent crs

resqpy/olio/wellspec_keywords.py
```diff
@@ -31,96 +32,98 @@
...
-wellspec_dtype['SECT'] = str  # todo: need to check type
-wellspec_dtype['FLOWSECT'] = str  # todo: need to check type
-wellspec_dtype['ZONE'] = int
-wellspec_dtype['GROUP'] = str
-wellspec_dtype['TEMP'] = float
-wellspec_dtype['IPTN'] = int
-wellspec_dtype['D'] = float
-wellspec_dtype['ND'] = str
-wellspec_dtype['DZ'] = float
+# yapf: disable
```
connortann, line 40: FYI @andy-beer - I thought here was a good place to try out the `# yapf: disable` to keep the previous indentation, seems to work as expected.

andy-beer: Yes, that looks good.

I think most of the data in wellspec_dict is only used by pagoda and probably should be stripped out of resqpy anyway!  However, I can look into that separately at a later date.

connortann (approved):
LGTM! @andy-beer  I fixed the remaining typing error by changing the order of arguments in the overwritten `create_xml` method, to ensure it is compatible with the base class method. Over to you to merge if happy.

Issue #131, opened 2021-07-06
Reduce complexity of largest functions: limit the Cyclomatic complexity


connortann:
Some functions and methods are complex with multiple nested control flow statements, making them difficult to read, understand or maintain. The most complex functions are currently: 

- `BlockedWell.dataframe`
- `transmission.fault_connection_set`
- `Grid.set_geometry_is_defined`.

The [Cyclomatic Complexity](https://en.wikipedia.org/wiki/Cyclomatic_complexity) is a measure of how complex a function is. It is the number of linearly independent possible paths in a function. It relates to how many test cases would be needed for complete logical coverage.

According to McCabe, anything that goes beyond 10 is too complex. Some functions in resqpy currently have a cyclomatic complexity of over 100, so it would be great to refactor them.

The Cyclomatic complexity can be calculated with static analysis, for example with the [McCabe flake8 plugin](https://github.com/PyCQA/mccabe). This could be set up on CI, to enforce an upper limit to complexity.

cflynn3:
This has been broken down into other issues, most of which have been completed or will be completed.

PR #132, opened 2021-07-06
Enable max-complexity CI checks

connortann:
Supports #131 , but does not fully close.

Implements a very generous cyclometric complexity limit, 5 times greater than the recommended limit.

This change makes our static analysis bot comment on any functions that are too complex, according to the cyclomatic complexity threshold set in `setup.cfg`. See the "files changed" tab:

image

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -57,6 +57,7 @@ filterwarnings =
 [flake8]
 # count = True
 # statistics = True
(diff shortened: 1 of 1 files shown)


connortann:
Closing for now - cannot enable complexity CI checks until existing complex functions are refactored.

PR #133, opened 2021-07-07
WIP: adds WELLSPEC column data retrieval from blocked well properties

andy-beer:
This change adds further integration of blocked well properties when reading or writing Nexus wellspec data (or working with the closely related dataframe).


andy-beer:
Note that these changes are an interim solution for handling wellspec data as properties.  The dataframe method in the BlockedWell class is likely to be refactored in the near future.

PR #142, opened 2021-07-09
Add Pillow as dependency

connortann:
I noticed `resqpy.tomography` imports `PIL`, which is not available in the project's dependencies. This will lead to an import error if the module is used in a fresh environment.

This PR tries to fix that by adding the library to the main list of dependencies. An alternative could be to add it as an optional extra, under "extras_requires".

It looks like the Python Imaging Library `PIL` is no longer maintained, and `Pillow` is the actively used fork, so suggesting putting that into our dependencies.

I've added `resqpy.tomography` to the TOC too, as presumably we want this to appear in the docs.

I believe it's not used in any unit tests, and is missing from the documentation table of contents, which is why the CI suite did not pick this up.

### DIFF
--- a/resqpy/__init__.py
+++ b/resqpy/__init__.py
@@ -18,7 +18,9 @@
     rq_import
     surface
     time_series
+    tomography
     well
+    weights_and_measures
     olio
 
 """
--- a/setup.cfg
+++ b/setup.cfg
@@ -30,6 +30,7 @@ install_requires =
     h5py
     lxml
     lasio
+    Pillow
 
 [options.package_data]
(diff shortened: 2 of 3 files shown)


andy-beer:
Alternatively, we could pull the tomography module out of resqpy and keep it as a bp internal module – I'm undecided what is best, let's discuss.

andy-beer (approved):
Okay, we might as well add that Pillow dependency and keep the tomography module in resqpy.

connortann:
As discussed - decided to move tomography module elsewhere for now.

PR #143, opened 2021-07-12
attempt to resolve issues with blocked well refinement

andy-beer:
This request is for changes needed to improve the intersection of simulation trajectories with a refined grid (LGR).

- commented out a couple of debug statements in base.py


andy-beer:
I'm merging these changes even though the basic issue is not resolved and cannot be without use of original trajectory data.

PR #146, opened 2021-07-13
adding more tests for some olio modules

andy-beer:
- box_utilities


tests/test_box_utilities.py
```diff
@@ -0,0 +1,82 @@
+# test olio box_utilities functions
+
+import pytest
+import numpy as np
+
+import resqpy.olio.box_utilities as bx
+
+
+def test_box_utilities():
```
connortann, line 9: I think this function would be a lot more understandable if it were broken out into individual unit tests with descriptive names. It is hard for me to understand what the intent of this test is from the code.

Rather than one 70-line test function with no commentary, could it be 10 7-line unit tests?

andy-beer: I've added comments for now. If you think breaking it into smaller functions is important I could do that, and simply have a common function for generating the tiny test data that is common to most of the tests. In my opinion these are quite simple tests of quite simple functionality, so I'm not sure it's worth putting much effort into.

connortann: Those comments help a lot, thanks!

connortann (approved):

PR #150, opened 2021-07-13
Relocate tomography module

connortann:


andy-beer (approved):
If we remove the tomography module from resqpy, I think we can remove the Pillow dependency.

connortann:
There is no Pillow dependency in resqpy. #142 was closed without merging after you suggested simply removing the module.

PR #151, opened 2021-07-13
Move release steps from README to CONTRIBUTING

connortann:


docs/CONTRIBUTING.rst
```diff
@@ -71,24 +81,17 @@ Development environment setup
...
-You should either set this formatter in your IDE settings, or run the following
-command before committing:
+Most IDEs allow you to configure a formatter to run automatically when you save
+a file. Alternatively, you can run the following command before commiting any
+changes:
 
 .. code-block:: bash
 
+   # Reformat all python files in the repository
    yapf -i .
```
andy-beer, line 95: I think we need -ir here

andy-beer: Yes, on my mac I get the following error with the command as currently written. I will make the change to this branch and then approve: I hope that works for you.

andy-beer (approved):

PR #154, opened 2021-07-14
replacing some secondary root reference attributes with uuids

andy-beer:
The main focus of changes are in the well module. However, the model and lines modules have also been modified in a similar way.


emmanesbit (approved):
Looks good to me

PR #162, opened 2021-07-20
Option to cut references to given uuids during copy

andy-beer:
adding option to cut references to given uuids during copy part from other model


emmanesbit (approved):
Looks good to me

PR #163, opened 2021-07-23
Gh155/rel perm

kadijahassanali:
Hello everyone,

I'd appreciate feedback on my attempt at writing a RelPerm class and the corresponding unit tests.

Thanks!

Best regards,
Kadija


kadijahassanali:
The failure in the Unit tests appears to be due module import errors for scipy and reqspy.olio.dataframe
I have no idea why the get_relperm_parts_in_model method is failing 🗡️ 👎

resqpy/olio/relperm.py
```diff
@@ -0,0 +1,311 @@
...
+"""
+import numpy as np
+import pandas as pd
+from scipy import interpolate
+import os
+import logging
+import resqpy.olio.xml_et as rqet
+from resqpy.olio.dataframe import DataFrame
+
+version = '6th May 2021'
```
emmanesbit, line 16: Update version

resqpy/olio/relperm.py
```diff
@@ -0,0 +1,311 @@
...
+           the low-salinity data for relative permeability and capillary pressure
+           table_index (int, optional): the index of the relative permeability
+           table when multiple relative permeability tables are present
+
+        note:
+           see DataFrame class docstring for details of other arguments
+        """
+
+        # check that either a uuid OR dataframe has been provided
+        assert uuid is not None or df is not None
```
emmanesbit, line 58: `, 'Either uuid or dataframe must be provided'`

resqpy/olio/relperm.py
```diff
@@ -0,0 +1,311 @@
...
+            uom_lookup_uuid=None):
+        """Create a new RelPerm object from either a previously stored property or a pandas dataframe.
+
+        arguments:
+           phase_combo (str, optional): the combination of phases whose relative
+           permeability behaviour is described. Options include 'water-oil', 'gas-oil' and
+           'gas-water'
+           low_sal (boolean, optional): if True, indicates that the water-oil table contains
+           the low-salinity data for relative permeability and capillary pressure
+           table_index (int, optional): the index of the relative permeability
```
emmanesbit, line 50: Might be worth mentioning these are indexed to 1, not 0

emmanesbit (commented):
Looks good to me - great set of tests to cover all the input options.

I'm not sure why the tests are failing now that scipy has been added in, will take another look at the errors now

resqpy/olio/relperm.py
```diff
@@ -0,0 +1,316 @@
...
+                df = None,
+                uom_list = None,
+                realization = None,
+                phase_combo = None,
+                low_sal = False,
+                table_index = None,
+                title = 'relperm_table',
+                column_lookup_uuid = None,
+                uom_lookup_uuid = None):
+      """Create a new RelPerm object from either a previously stored property or a pandas dataframe.
```
andy-beer, line 41: 'property' has a specific technical meaning in RESQML, probably better to use another word such as 'object' here.

resqpy/olio/relperm.py
```diff
@@ -0,0 +1,316 @@
...
+               df[col].replace('None', np.nan, inplace = True)
+
+         # convert all values in the dataframe to numeric type
+         df_cols = df.columns
+         df[df_cols] = df[df_cols].apply(pd.to_numeric, errors = 'coerce')
+
+         # ensure that no other column besides Pc has missing values
+         cols_no_pc = [x for x in df.columns if 'pc' != x.lower()]
+         for col in cols_no_pc:
+            if col.capitalize == 'Pc':
```
andy-beer, line 120: Ye of little faith!  No need to have this as well as the cols_no_pc setup: I suggest using just one or the other

resqpy/olio/relperm.py
```diff
@@ -0,0 +1,316 @@
...
+   def interpolate_point(self, saturation, kr_or_pc_col):
+      """Returns a tuple of the saturation value and the corresponding
+        interpolated relative permeability or capillary pressure value.
+
+        arguments:
+            saturation (float): the saturation at which the relative permeability or cap. pressure will be interpolated
+            kr_or_pc_col (str): the column name of the parameter to be interpolated
+        returns:
+            tuple of float, the first element is the saturation and the second element is the interpolated value
+           """
```
andy-beer, line 162: Could be worth adding a note to say that a simple linear interpolation is used.

resqpy/olio/relperm.py
```diff
@@ -0,0 +1,316 @@
...
+
+   def df_to_text(self, filepath, filename):
+      """Creates text file from a dataframe of relative permeability and capillary pressure data.
+
+        arguments:
+            filepath (str): location where new text file is written to
+            filename (str): name of the new text file
+        returns:
+            tuple of float, the first element is the saturation and the second element is the interpolated value
+           """
```
andy-beer, line 185: I suggest a note saying that the format is for Nexus (though other formats could be supported in future).
If the word Nexus does appear anywhere, please add a trademark comment at the top of the module, something like:
# Nexus is a registered trademark of Halliburton

resqpy/olio/relperm.py
```diff
@@ -0,0 +1,316 @@
...
+   """
+    Returns a dictionary that contains dataframes with relative permeability and capillary pressure data and
+    phase combinations.
+
+    arguments:
+    filepath (str): relative or full path of the text file to be processed
+
+    returns:
+    dict, each element in the dictionary contains a dataframe, with saturation and rel. permeability/capillary pressure
+    data, and the phase combination being described
```
andy-beer, line 241: Again I suggest a comment saying that Nexus format is currently used

resqpy/olio/relperm.py
```diff
@@ -0,0 +1,316 @@
...
+                       support_root = None,
+                       df = df,
+                       uom_list = uom_list,
+                       realization = realization,
+                       title = title,
+                       column_lookup_uuid = column_lookup_uuid,
+                       uom_lookup_uuid = uom_lookup_uuid)
+      self.phase_combo = phase_combo
+      self.low_sal = low_sal
+      self.table_index = table_index
```
andy-beer, line 151: I think we need to save these 3 extra attributes as extra metadata items. Then when loading from a uuid, set (and/or check) the values based on that extra metadata.

resqpy/olio/relperm.py
```diff
@@ -0,0 +1,316 @@
...
+   def write_hdf5_and_create_xml(self):
+      """Write relative permeability table data to hdf5 file and create xml for RESQML
+        objects to represent dataframe."""
+
+      super().write_hdf5_and_create_xml()
+      # note: time series xml must be created before calling this method
+      mesh_root = self.mesh.root
+      # create an xml of extra metadata to indicate that this is a relative permeability table
+      rqet.create_metadata_xml(mesh_root, {'relperm_table': 'true'})
+
```
andy-beer, line 229: I believe we should add each of the three extra attributes as extra metadata items (as well as the relperm_table item):

- phase_combo
- low_sal
- table_index

resqpy/olio/relperm.py
```diff
@@ -0,0 +1,316 @@
...
+      df = pd.DataFrame(table_rows, columns = table_cols)
+      df.columns = df.columns.map(df_cols_dict)
+      sat_col = [x for x in df.columns if 'S' in x][0]
+      df = df.apply(pd.to_numeric, errors = 'coerce')
+      df = df[df[sat_col].notnull()]
+      rel_perm_dict[key]['df'] = df
+   return rel_perm_dict
+
+
+def relperm_parts_in_model(model, title = None, related_uuid = None):
```
andy-beer, line 297: If the 3 extra items are added as extra metadata, they could also be optional arguments to this method, so calling code could say (for example) _low_sal = True_ to limit the returned parts to low salinity tables. The default of None would need to cause that item to be left out of the dictionary passed to model.parts()

andy-beer (changes requested):
This is a very nice piece of coding Kadija!

Could we discuss the extra metadata items which I have suggested, please.  Otherwise good to merge from my side.

resqpy/olio/relperm.py
connortann: `str.capitalize` is a method, so this would need brackets `()`

resqpy/olio/relperm.py
```diff
@@ -0,0 +1,316 @@
+"""relperm.py: class for storing and retrieving dataframes of relative permeability
+data as RESQML objects.
+
+   note that this module uses the obj_Grid2dRepresentation class in a way that was not envisaged
+   when the RESQML standard was defined; software that does not use resqpy is unlikely to be
+   able to do much with data stored in this way
+"""
+import numpy as np
+import pandas as pd
+from scipy import interpolate
```
connortann, line 10: To avoid adding extra dependencies to resqpy, could we use `numpy.interp` rather than `scipy.interpolate`?

resqpy/olio/relperm.py
```diff
@@ -0,0 +1,316 @@
...
+                realization = None,
+                phase_combo = None,
+                low_sal = False,
+                table_index = None,
+                title = 'relperm_table',
+                column_lookup_uuid = None,
+                uom_lookup_uuid = None):
+      """Create a new RelPerm object from either a previously stored property or a pandas dataframe.
+
+        arguments:
```
connortann, line 43: Some arguments of the init method are missing from this docstring

resqpy/olio/relperm.py
```diff
@@ -0,0 +1,316 @@
...
+           the low-salinity data for relative permeability and capillary pressure
+           table_index (int, optional): the index of the relative permeability
+           table when multiple relative permeability tables are present. Note, indices should start at 1.
+
+        note:
+           see DataFrame class docstring for details of other arguments
+        """
+
+      # check that either a uuid OR dataframe has been provided
+      assert uuid is not None or df is not None, 'either a uuid or a dataframe must be provided'
```
connortann, line 57: A minor suggestion - I think it would be more appropriate to raise a ValueError rather than an AssertionError in cases like this where the user-provided inputs are "wrong". Same for the other validation checks below.

`assert` is generally used to check for "internal" consistency, for ease of debugging and maintenance. AssertionErrors should generally not be thrown to a user unless there is a mistake in resqpy.

resqpy/olio/relperm.py
```diff
@@ -0,0 +1,316 @@
...
+      # check that 'phase_combo' parameter is valid
+      processed_phase_combo = set([x.strip() for x in str(phase_combo).split('-')])
+      assert processed_phase_combo in [{'water', 'oil'}, {'gas', 'oil'}, {'gas', 'water'},
+                                       {'None'}], 'invalid phase_combo provided'
+
+      # check that table_index is >= 1
+      if table_index is not None:
+         assert table_index >= 1, 'table_index cannot be less than 1'
+
+      # check that the column names and order are as expected
```
connortann, line 68: I really like this use of comments at the start of each block of code - it makes the code very readable.

resqpy/olio/relperm.py
```diff
@@ -0,0 +1,316 @@
...
+                       title = title,
+                       column_lookup_uuid = column_lookup_uuid,
+                       uom_lookup_uuid = uom_lookup_uuid)
+      self.phase_combo = phase_combo
+      self.low_sal = low_sal
+      self.table_index = table_index
+
+   def interpolate_point(self, saturation, kr_or_pc_col):
+      """Returns a tuple of the saturation value and the corresponding
+        interpolated relative permeability or capillary pressure value.
```
connortann, line 155: The docstrings need to have perfectly aligned indentation for the built documentation to work.

Docstrings should start with a single one-line summary, followed by a blank line. The subsequent items (e.g. "arguments", "returns", or the closing triple quotes) should be at the same level of indentation as the opening triple quotes.

resqpy/olio/relperm.py
```diff
@@ -0,0 +1,316 @@
...
+      rqet.create_metadata_xml(mesh_root, {'relperm_table': 'true'})
+
+
+def text_to_relperm_dict(filepath):
+   """
+    Returns a dictionary that contains dataframes with relative permeability and capillary pressure data and
+    phase combinations.
+
+    arguments:
+    filepath (str): relative or full path of the text file to be processed
```
connortann, line 237: I think the indentation in this docstring needs to be fixed: "arguments" would be one indent, "filepath" would be two indents.

With resqpy's unusual 3-space indentation this might be difficult depending on your editor settings!

tests/test_relperm.py
```diff
@@ -0,0 +1,136 @@
+# test module for the derived RelPerm class located in resqpy.olio.dataframe.py
+
+import os
+import shutil
+import tempfile
+import unittest
```
connortann, line 6: Resqpy currently uses the high-level `pytest` testing framework rather than `unittest`, so for consistency I would suggest switching to pytest.

The syntax is similar:

- The individual unit tests would simply be functions rather than methods within a class.
- Fixtures would replace the "setUp" and "tearDown" methods. You could use the [tmp_path](https://docs.pytest.org/en/6.2.x/tmpdir.html) fixture to get a temporary directory, or the `tmp_model` fixture (which defined in `resqpy/tests/conftest.py`) to get a fresh new model.
- The regular `assert` replaces `self.assertRaises`
- `pytest.raises` replaces `self.assertRaises`
- One does not need the final two lines (unittest.main())

connortann (changes requested):
Suggested some changes. Code overall looks very well-written, and great to see unit tests and documentation!

kadijahassanali:
Thanks for the feedback @andy-beer and @connortann! 

I'll spend today making those requested changes and reach out if I have any questions 👍.

resqpy/olio/relperm.py
kadijahassanali: Done!

resqpy/olio/relperm.py
kadijahassanali: Gotcha....I changed all the assert statements in the validation checks to if-else statements that raise ValueErrors

tests/test_relperm.py
kadijahassanali: I refactored the unit tests to use the pytest framework.

resqpy/olio/relperm.py
kadijahassanali: Done!

resqpy/olio/relperm.py
kadijahassanali: Done! I've added the trademark comment to the top of the relperm.py file

resqpy/olio/relperm.py
kadijahassanali: I added a note to the method's docstring to indicate this.

resqpy/olio/relperm.py
kadijahassanali: I think I fixed the docstring issues based on your previous advice....maybe

resqpy/olio/relperm.py
kadijahassanali: Hi Andy, 
I've gone ahead and saved the attributes as extra_metadata items in the RelPerm class's init method.

resqpy/olio/relperm.py
kadijahassanali: I've now included these attributes as extra_metadata items.

resqpy/olio/relperm.py
kadijahassanali: The method now includes the option to narrow the search based on these 3 extra metadata items :)

connortann:
@kadijahassanali  FYI, it appeared that there was a huge diff in `olio.dataframes`, despite the fact that very few lines were actually changed. I believe this is because the line endings were changed (from linux-style to windows-style). This shouldn't normally happen, unless you have unusual git settings or a unusual workflow for editing and saving files.

In any case, I've added a ".gitattributes" file to tell git explicitly to handle line endings in a consistent fashion, and pushed a commit to revert the line endings to how they were. The diff now looks sensible again.

PR #164, opened 2021-08-04
Bug fixes to lines, and additional unit tests

emmanesbit:
Fixing errors in Polylines and PolylineSet following changes to crs_root in previous MRs. Added unit tests for Polylines, PolylineSets to ensure generated lines can be reloaded, and added tests for charisma and irap file loading.


connortann:
@emmanesbit  this looks great! Love the example_data_path fixture :D

One thing jumps out to me - I think the init method of `PolylineSet` is doing a fair bit of work and could be refactored. In other resqpy classes, we've tried separate out any alternative constructor options as separate class methods. An example is `resqpy.well.DeviationSurvey.from_data_frame` . This helps keep the init method simple.

I would suggest making class methods e.g.`PolylineSet.from_irap_file()` and `PolylineSet.from_charisma_file()`

tests/test_lines.py
```diff
@@ -25,8 +28,80 @@ def test_lines(example_model_and_crs):
...
+   assert (reload.count_perpol == [
+      4, 5, 4, 5, 5
+   ]).all(), f"Expected count per polyline to be [4,5,4,5,5], found {reload.count_perpol}"
+   assert len(reload.coordinates) == 23, f"Expected length of coordinates to be 23, found {len(reload.coordinates)}"
+
+
+def test_irap(example_model_and_crs, example_data_path):
+   # Set up a PolylineSet
+   model, crs = example_model_and_crs
+   irap_file = os.path.join(example_data_path, "IRAP_example")
```
connortann, line 96: A minor comment: when using Pathlib objects rather than strings there is a neat high-level syntax for path operations. One can simply write `irap_file = example_data_path / "IRAP_example"`

P.S., should those files have a file extension, like ".txt"?

connortann (changes requested):
Suggest splitting out constructor methods as mentioned above

tests/test_lines.py
andy-beer: Oddly enough I dislike the overloading of / for path concatenation. However, listen to @connortann , not me on this one!

connortann: Either is fine I think.

However - we should probably make sure that all user-facing resqpy code that accepts filepaths is able to handle `pathlib.Path` objects as well as simple strings. `pathlib` (the higher-level builtin python lib for file paths) is quite widely used these days rather than `os.path`.

Unless there are objections, I'll open a separate issue that that one and work on it.

tests/conftest.py
```diff
@@ -96,3 +99,18 @@ def example_model_with_logs(example_model_with_well):
...
    return model, well_interp, datum, traj, frame, log_collection
+
+
+@pytest.fixture
+def example_data_path(tmp_path):
+   """ Return pathlib.Path pointing to temporary copy of tests/example_data
+
+   Use a fresh temporary directory for each test.
+   """
+   master_path = (Path(__file__) / '../example_data').resolve()
```
andy-beer, line 110: We currently have two directories called example_data: one at the top level for example RESQML datasets for people to play with, and one within the test directory.  I suggest we rename the one in the test directory as test_data. (Doesn't need to be part of this change but I thought I'd mention it here before forgetting about it again!)

andy-beer (approved):
Okay to merge as is, from my side.

One other random thought: Could you please check the UK registered trademarks database, @emmanesbit , to see if Charisma, irap or GOCAD should be listed as trademarks in the resqpy documentation (and in comments at the top of relevant modules). In the past I have used the UK government search site: https://www.gov.uk/search-for-trademark

emmanesbit:
> Okay to merge as is, from my side.
> 
> One other random thought: Could you please check the UK registered trademarks database, @emmanesbit , to see if Charisma, irap or GOCAD should be listed as trademarks in the resqpy documentation (and in comments at the top of relevant modules). In the past I have used the UK government search site: https://www.gov.uk/search-for-trademark

@andy-beer I can't see anything that looks like it relates to the data types I refer to here for either of the three.

tests/conftest.py
emmanesbit: @andy-beer  @connortann  - I did notice there were two, though the example_data folder within the test directory contains a couple of extra files. I'm not sure if/where these files are used - does anyone know where these are used so I can update the references to these at the same time? If not I might break something when renaming

connortann: I think you're fine to go ahead and rename the tests/example_data directory. I think I added that LAS file for a test of `resqpy.well.add_las_to_trajectory`, but haven't yet got round to writing that test!

connortann (approved):
I've moved the class method idea to a new issue #167

PR #170, opened 2021-08-04
Remove duplicate test data, make relperm test more forgiving

connortann:
A bunch of changes to clean up the test suite:

1. The test suite was passing on CI but failing on my local machine. Fixes that by using more permissive `assert_dataframe_equal`
2. Deletes the redundant tests/example_data directory
3. A large number of small changes to fix Deprecation warnings in the test suite, supporting #17


tests/test_relperm.py
```diff
@@ -125,7 +126,7 @@ def test_relperm(tmp_path):
    assert len(relperm_parts_in_model(model, low_sal = False)) == 1
    dataframe1.df_to_text(filepath = tmp_path, filename = 'oil_water_test_table')
    df1_reconstructed = text_to_relperm_dict(os.path.join(tmp_path, 'oil_water_test_table.dat'))['relperm_table1']['df']
-   assert df1.equals(df1_reconstructed)
+   assert_frame_equal(df1, df1_reconstructed)
```
kadijahassanali, line 129: This is such a useful function...thanks for sharing Connor. The changes look good.

resqpy/model.py
```diff
@@ -3274,11 +3274,9 @@ def iter_trajectories(self):
...
       import resqpy.well  # Imported here for speed, module is not always needed
 
-      parts = self.parts_list_of_type('WellboreTrajectoryRepresentation')
-      for part in parts:
-         traj_root = self.root_for_part(part)
-         traj = resqpy.well.Trajectory(self, trajectory_root = traj_root)
-         yield traj
+      uuids = self.uuids(obj_type = "WellboreTrajectoryRepresentation")
+      for uuid in uuids:
+         yield resqpy.well.Trajectory(self, uuid = uuid)
```
kadijahassanali, line 3279: Out of curiosity, why are we switching from using parts to uuids?

kadijahassanali (approved):

resqpy/model.py
connortann: This relates to issue #17 , our longer-term goal to clean up the code base. Presently, uuids roots and part names are used interchangably as the "key" to reference objects, which leads to a fair amout of messy code having to covert between them all. We decided to standardise on `uuids`, as use that as the only "key" if possible throughout the codebase.

Most classes now support an init method with a `uuid` parameter.

PR #182, opened 2021-08-19
Add extra_metadata to pointset

emmanesbit:
Adding extra_metadata saving and reading to PointSetRepresentation

### DIFF
--- a/resqpy/surface.py
+++ b/resqpy/surface.py
@@ -1198,7 +1198,8 @@ class PointSet(_BaseSurface):
                        uuid = uuid,
                        title = title,
                        originator = originator,
-                       root_node = point_set_root)
+                       root_node = point_set_root,
+                       extra_metadata = extra_metadata)
 
       if self.root is not None:
          if load_hdf5:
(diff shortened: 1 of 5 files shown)


resqpy/surface.py
```diff
@@ -1534,6 +1535,14 @@ def convert_to_irap(self, file_name):
          for item in lines:
             f.write(item)
 
+   def append_extra_metadata(self, meta_dict):
```
connortann, line 1538: Would this make sense as a general method of the base class `BaseResqpy`, or is it PointSet specific? Also, perhaps the word "extra" in the title is redundant?

resqpy/surface.py
```diff
@@ -1534,6 +1535,14 @@ def convert_to_irap(self, file_name):
...
+      """Append a given dictionary of metadata to the existing metadata."""
+
+      if self.extra_metadata is None:
+         self.extra_metadata = {}
+      for key in meta_dict:
+         self.extra_metadata[key] = meta_dict[key]
```
connortann, line 1544: 1. Would be great to have a unit test
2. Is it possible for extra_metadata to be None? I think it might be an always present dictionary
3. To merge dicts in python 3.5+ one can use `z = {**x, **y}`. But either way is fine, explicit is good too I think!

connortann (approved):
LGTM

PR #191, opened 2021-08-20
Minor changes to read and apply mults from ascii

emmanesbit:
Updates to fault.py add_connection_set_and_tmults to read in the mult values from the ascii and apply the tmult_dict to these values. Previous approach assumed these values were always 1.


emmanesbit:
Additional unit tests added. And additional conftest with example_model_with_properties and grid.

resqpy/fault.py
```diff
@@ -1614,28 +1614,18 @@ def k_gap_connection_set(grid, skip_inactive = True, feature_name = 'k gap conne
    return kgcs
 
 
-def add_connection_set_and_tmults(model, fault_incl, tmult_dict):
+def add_connection_set_and_tmults(model, fault_incl, tmult_dict = {}):
```
connortann, line 1617: We should avoid having mutable default arguments (as per one of the recent Bifrost Python Tip threads).

A common solution is having a default of `None`, and a clause like `if tmult_dict is None: dmult_dict = {}`

resqpy/fault.py
```diff
@@ -1614,28 +1614,18 @@ def k_gap_connection_set(grid, skip_inactive = True, feature_name = 'k gap conne
...
    """Add a grid connection set to a resqml model, based on a fault include file and a dictionary of fault:tmult pairs.
 
    Grid connection set added to resqml model, with extra_metadata on the fault interpretation containing the MULTFL values
 
    Args:
       model - resqml model object
       fault_incl - fullpath to fault include file or list of fullpaths to fault include files
-      tmult_dict - dictionary of fault name: tmult pairs (must align with faults in include file)
+      tmult_dict - dictionary of fault name: tmult pairs (must align with faults in include file). Optional, if blank values in the fault.include file will be used instead
```
connortann, line 1625: For docs to render nicely, Sphinx expects arguments to be listed with colons rather than dashes. Would also suggest a line break:

```suggestion
      tmult_dict: dictionary of fault name: tmult pairs (must align with faults in include file).
          Optional, if blank values in the fault.include file will be used instead
```

connortann: Also, it could be helpful to document here what a tmult pair is. I'm guessing a pair of transmissibility multipliers?

connortann (changes requested):
Two minor suggestions but otherwise looks great

connortann (approved):
LGTM!

Issue #197, opened 2021-08-27
Question: how to create an epc of properties without including the grid


mgimhof:
I have an application that creates an EPC file containing only ContinuousProperties and the EpcExternalPartReference to the h5 file.  The use case is python as an external property calculator. The grid is not explicitly included because the app already contains the grid and it does not need to be round tripped (it is a very simple-minded calculator that does not account for grid geometry and topology!).

I can pick up properties and perform some calculation to create a new one. The problem is when trying to create an epc for the new property.

Happening to know the number of cells and the grid uuid, I expected that the return problem can be mocked up as

```
import numpy as np

import resqpy.model as rq
import resqpy.property as rp

CONTENT_TYPE = 'obj_IjkGridRepresentation'
SUPPORT_UUID = '492f069b-888a-4a17-8cc1-cdc845774f18'
NUMCELLS = 64

new_model = rq.Model('d:/new.epc', new_epc=True)
grid = new_model.create_supporting_representation(support_uuid=SUPPORT_UUID, title='grid',
                                                  content_type=CONTENT_TYPE)
new_model.add_part(content_type=CONTENT_TYPE, uuid=SUPPORT_UUID, root=grid)
values = np.arange(start=0, stop=NUMCELLS, dtype=np.float32)
p = rp.Property.from_array(new_model, cached_array=values, source_info='rpm', keyword='PERM_IJ',
                           property_kind='permeability rock', support_uuid=SUPPORT_UUID)

new_model.store_epc()
```

If the add_part line is included, then I fail with 
```
C:\Users\m9imhof\AppData\Local\ExxonMobil\PetrelPythonPlugin\pytrel\python.exe D:/DEV/resqpy/test_local/import_export.py
Traceback (most recent call last):
  File "D:\DEV\resqpy\test_local\import_export.py", line 15, in <module>
    p = rp.Property.from_array(new_model, cached_array=values, source_info='rpm', keyword='PERM_IJ',
  File "D:\DEV\resqpy\resqpy\property.py", line 3542, in from_array
    prop.create_xml(support_uuid = support_uuid,
  File "D:\DEV\resqpy\resqpy\property.py", line 3736, in create_xml
    self.collection.create_xml_for_imported_list_and_add_parts_to_model(
  File "D:\DEV\resqpy\resqpy\property.py", line 2685, in create_xml_for_imported_list_and_add_parts_to_model
    p_node = self.create_xml(
[...]

PR #200, opened 2021-08-30
new method in Model for temp supporting representation

andy-beer:
This change is to provide method(s) allowing the creation of a temporary (mock) supporting representation, when working with a dataset where the actual supporting representation is not included.

This relates to issue #197


andy-beer:
The new Model method change_uuid_in_supporting_representation_reference() allows patching of the uuid of the supporting representation in the xml reference block (within a property object). It does not modify _rels information.

PR #214, opened 2021-09-10
adds support for 'nodes' and 'nodes per cell' indexable elements; closes #211

### LINKED ISSUE #211: Add 'nodes' as supported indexable element for IJK grid properties
The PropertyCollection class does not currently handle 'nodes' as an indexable element for grid properties. This capability should be added for IJK grid properties.

Note that the shape of the points array (nodes) is different depending on whether an IJK grid is faulted (has split pillars) or not.  Calling code would have to ensure that properties assigned to nodes are correctly set up, which is non-trivial when split pillars are present.

andy-beer:
adds support for 'nodes' and 'nodes per cell' indexable elements for IJK grid properties

resolves issue #211

### DIFF
--- a/tests/test_property.py
+++ b/tests/test_property.py
@@ -31,6 +31,24 @@ def test_property(tmp_path):
                                 property_kind = 'net to gross ratio',
                                 indexable_element = 'cells',
                                 uom = 'm3/m3')
+   a3 = np.random.random((grid.nk + 1, grid.nj + 1, grid.ni + 1))
+   p3 = rqp.Property.from_array(model,
+                                a3,
+                                source_info = 'random',
+                                keyword = 'jiggle shared nodes',
+                                support_uuid = grid.uuid,
+                                property_kind = 'length',
+                                indexable_element = 'nodes',
+                                uom = 'm3')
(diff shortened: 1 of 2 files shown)


tests/test_property.py
```diff
@@ -31,6 +31,24 @@ def test_property(tmp_path):
...
+   p3 = rqp.Property.from_array(model,
+                                a3,
+                                source_info = 'random',
+                                keyword = 'jiggle shared nodes',
+                                support_uuid = grid.uuid,
+                                property_kind = 'length',
+                                indexable_element = 'nodes',
+                                uom = 'm3')
+   a4 = np.random.random((grid.nk, grid.nj, grid.ni, 2, 2, 2))
+   p3 = rqp.Property.from_array(model,
```
emmanesbit, line 44: Should this be p4?

andy-beer: Well spotted @emmanesbit !  I've pushed a correction.  (As it happens, that variable is not used afterwards in the test, so no change to behaviour.)

emmanesbit:
Looks good to me!

PR #216, opened 2021-09-12
uom aliases added for some units used in Nexus output

andy-beer:
This change adds some more unit aliases, as used in some Nexus output.

Nexus is a trademark of the Halliburton Company.


connortann (approved):
LGTM

PR #220, opened 2021-09-15
suppressing property kind warning for length-thickness mismatches

andy-beer:
This change suppresses the property kind warning when the kind and expected kind are both one of 'length' or 'thickness'.


emmanesbit (approved):
Looks good

PR #221, opened 2021-09-15
module version string updates

andy-beer:


andy-beer:
This merge was mostly to encourage a regeneration of the readthedocs which seemed to have become corrupted.

Issue #223, opened 2021-09-15
requirement to populate a Grid's geometry from a points property


andy-beer:
The Grid class includes many methods which work with the geometry of the grid, as stored in the points array. This issue is to allow a means of populating the points data from a points property (for a given realization and/or time index as applicable).

The resqpy Grid object would effectively have a dynamic geometry, even though the official geometry (if any) stored to RESQML would be static.

Re-populating the points would need various caches (such as centre point, volume etc.) to be invalidated.

The initial use case is to support basin modelling.

PR #226, opened 2021-09-15
setting grid geometry from points property; resolves issue #223

andy-beer:
This change provides a Grid method for overwriting the cached points (geometry) from an instance of a points property.  The property may have multiple realizations and/or time indices as usual.  This is primarily to support basin modelling or other geological processes affecting the geometry of strata.


andy-beer:
Initial code written.  Not sure how to handle the grid's inactive attribute.

PR #235, opened 2021-09-21
Add context manager for Models

connortann:
Adds a high-level way of opening and closing models:

```python
with ModelContext("my_model.epc") as model:
    print(model.uuids())
```

This provides a convenient way to ensure all file handles are properly closed when the "with" clause exists


connortann:
FYI @jrt54 , how does this look as a context manager?

andy-beer:
I favour keeping this functionality separate from the Model class, ie. as currently implemented here.  A few other thoughts:

- Instead of (or as well as) the _write_ boolean argument, it might be better to have a _mode_ argument ('r', 'w', or 'a'; or their verbose equivalents).  In the case of 'w', a new_model() call could be made on entry.  The h5_release() should be called on exit in all cases.  The distinction between reading, writing and appending could potentially come in handy in some other work around 'locking' a RESQML dataset whilst it is in use.  Perhaps we could have a discussion about this @connortann ?
- The Model class includes a _modified_ attribute which can, in principal, be used to test whether a store_epc() is needed or not.
- Sometimes it is desirable to create resqpy objects and temporarily add them to a Model object without any intention of writing them to persistent storage, so control over the store_epc() does need to be available to calling code.
- Partly because of the possible use of temporary resqpy objects, trying to automate writing to hdf5 would need a lot of careful design – not recommended.

connortann:
Thanks for the feedback Andy! I'll plan to implement your suggestion of using a "mode" argument. There might possibly be confusion over "w" vs "a": I'm not sure what "append" would mean, as there isn't really a notion of where a file stream is positioned.

Perhaps the modes could be something like:

|        | if epc exists              | if epc doesn't exist | on exit      |
|--------|----------------------------|----------------------|--------------|
| read   | open it           | raise error          | close file handles           |
| write  | open it | new_model()     | save changes |
| create | raise error                | create new model     | save changes |

connortann:
@andy-beer  I've made some changes following your suggestions.

On an unrelated note - noticed we're now about 50% code coverage!!! Amazing!!!

andy-beer (approved):
Sorry for the delay in reviewing Connor!  This looks good to me – okay to merge.  You might consider making an edit to a tutorial sometime. (I don't think we need to change all the examples but this context manager should become the preferred way of working.)

tests/test_model.py
```diff
@@ -237,3 +237,40 @@ def test_model_copy_all_parts_non_resqpy_hdf5_paths(example_model_with_propertie
...
+   # Re-open model in read/write mode
+   with rq.ModelContext(epc_path, mode = "rw") as model2:
+
+      crs2 = rqc.Crs(model2, uuid = crs_uuid)
+      assert len(list(model2.iter_crs())) == 1
+      assert crs2.title == 'kuzcotopia'
+
+      # Make a change
+      crs2.title = 'wabajam'
+      crs2.create_xml(reuse = False)
```
connortann, line 262: @andy-beer In writing this unit test, a resqpy puzzle came up. It wasn't immediately obvious to me the correct way to make a change to an existing RESQML object, and save the change to disk.

In this test for example, I wanted to change the name of an existing CRS. If I call `create_xml()`, by default the "reuse" logic kicks in and the changes are not actually saved. If I set `reuse=False`, I think a second set of XML elements are created, but I don't think the first set are removed.

Do we need a "update_xml" method, or an "overwrite=True" argument, or something like that?

PR #251, opened 2021-10-26
very large tutorial page added with tables of RESQML class attributes

andy-beer:
I'm not sure this is a good idea, due to the size of the page – any thoughts on how this information should best be presented?


andy-beer:
I've added /docs/scripts/schema.py

That file is geared up to be used as a module for programmatically interrogating the RESQML schema, or to be run as a script to generate a single text file holding rst-format tables showing the elements of each obj_* class, as used in the tutorial page.

PR #257, opened 2021-11-04
Integration - surface.py - Move classes to individual files in new surface directory

emmanesbit:
Initial changes to break out classes from surface.py into individual files, under the directory 'surface'. And additional unit/integration tests for PointSet class.


connortann:
A quick suggestion to ensure that Sphinx continues to document things correctly: I think there needs to be an `__all__` statement in the `__init__.py`:

```python
"""The __init__ docstr"""

__all__ = ['func1', 'func2']

from m1 import func1
from m2 import func2
```

The `__all__` variable is sometimes used to designate what are the "publically exposed" classes and methods in a given module, and sphinx will use it to know what to document

connortann:
PS - I think you may need to re-run the yapf formatter, including on the test functions

emmanesbit:
> PS - I think you may need to re-run the yapf formatter, including on the test functions. If you're using VS code, that can be set in your project settings as an automatically-applied formatter that runs when you hit "save", which is a big time-saver! I can share more info if interested.

Ah, I was running it on the resqpy level below tests, so had no idea why it was going wrong! Thanks for the hint (and apologies if you get all the spam emails every time it fails!)

emmanesbit:
@cflynn3 - This is now ready for review, ahead of merging into the integration branch

Key changes:
- Break out surface.py file into surface directory containing several separate files, one for each class

- Additional integration tests for pointset.py
- Refactoring of pointset.py to simplify methods and reduce complexity
- Additional unit tests for pointset.py

Updates to other classes within the new surface directory will form separate pull requests

cflynn3 (approved):
These changes look great. Only small thing I'll say is I'd like to clearly separate the unit tests and integration tests, so perhaps have a 'unit tests' folder within the tests folder for files like test_surface_pointset.py?

kadijahassanali (approved):
Hi Emma, 

This looks great! I especially appreciated your unit and integration tests as they helped me understand the difference between the two. I will definitely be using the "Arrange", "Act", "Assert" methodology!

connortann:
I thought I share our offline conversation here for visibility and so others can chip in.

I propose we use **class constructor methods** for the various methods in PointSet that relate to loading from different file formats. I've opened a new issue #278 with more info and examples, to track this refactoring effort across all classes.

PR #279, opened 2021-11-10
Integration - surface/triangulated_patch.py - Unit tests and refactoring for TriangulatedPatch class

emmanesbit:


connortann (approved):
LGTM

emmanesbit:
@cflynn3 @kadijahassanali - this is now ready for review. Additional tests added to cover most functionality in TriangulatedPatch, and some refactoring to reduce complexity in some of the longer methods.

Note: set_to_sail and face_from_triangle_index both have no coverage currently. I'll take a closer look at these next week to understand these more.
- I can't quite get my head around what set_to_sail is doing, so reluctant to make any changes
- I'm not sure if there is a bug in face_from_triangle_index, or if I'm missing something here

resqpy/surface/triangulated_patch.py
```diff
@@ -307,6 +298,22 @@ def set_from_sparse_mesh(self, mesh_xyz):
         self.triangles = triangles[:nt, :]
         self.triangle_count = nt
 
+    def get_indices_from_sparse_meshxyz(self, mesh_xyz):
```
cflynn3, line 301: These new methods can be made private by adding a double underscore before the start of the name.

cflynn3 (approved):
Great work here, with one minor suggestion for improvement.

resqpy/surface/triangulated_patch.py
connortann: Agree that we should clearly delineate the public and non-public API.

PEP8 says "Use one leading underscore only for non-public methods and instance variables":

https://www.python.org/dev/peps/pep-0008/#method-names-and-instance-variables

kadijahassanali (approved):

PR #280, opened 2021-11-10
WIP: Merging the refactoring and testing improvements to RESQPY into master.

cflynn3:


resqpy/fault.py
```diff
@@ -20,7 +20,7 @@
 import resqpy.olio.uuid as bu
 import resqpy.olio.write_hdf5 as rwh5
 import resqpy.olio.xml_et as rqet
-import resqpy.organize as rqo
+import resqpy.organize.organize_old as rqo
```
connortann, line 23: @cflynn3 is this intentional? Doesn't look like a change that is ready to be accepted, but it has already been merged into the integration branch.

resqpy/organize/__init__.py
```diff
@@ -0,0 +1,6 @@
+__all__ = [
+    'OrganizationFeature',
+]
```
connortann, line 3: This is incomplete

resqpy/organize/organize_old.py
```diff
@@ -82,7 +81,7 @@ def create_xml_has_occurred_during(model, parent_node, hod_pair, tag = 'HasOccur
     model.create_ref_node('ChronoTop', model.title_for_root(chrono_top_root), top_chrono_uuid, root = hod_node)
 
 
-def _alias_for_attribute(attribute_name):
+def alias_for_attribute(attribute_name):
```
connortann, line 84: Should this be part of the public API? It is surely an implementation detail for attributes, hard to see why this should be exposed

resqpy/organize/organize_old.py
```diff
@@ -94,102 +93,101 @@ def fset(self, value):
...
+#             return True
+#         if check_extra_metadata and not equivalent_extra_metadata(self, other):
+#             return False
+#         return self.feature_name == other.feature_name
+#
+#     def create_xml(self, add_as_part = True, originator = None, reuse = True):
+#         """Creates a geobody feature xml node from this geobody feature object."""
+#         if reuse and self.try_reuse():
+#             return self.root  # check for reusable (equivalent) object
+#         return super().create_xml(add_as_part = add_as_part, originator = originator)
```
connortann, line 183: We probably do not want these comments included in our code

resqpy/organize/organize_old.py
cflynn3: This method is already being called from other modules (e.g. well.py), therefore the renaming is correct.

resqpy/grid.py
```diff
@@ -321,8 +319,7 @@ def denaturalized_cell_index(self, c0):
         return (k0, j0, i0)
 
     def denaturalized_cell_indices(self, c0s):
-        """Returns an integer numpy array of shape (..., 3) holding kji0 indices for the cells with given natural
-        indices.
+        """Returns an integer array holding kji0 indices for the cells with given natural indices.
 
         argument:
            c0s: numpy integer array of shape (...,) being natural cell indices (for a flattened array)
```
connortann, line 325: Think the "3" is missing here
```suggestion
           c0s: numpy integer array of shape (..., 3) being natural cell indices (for a flattened array)
```

resqpy/grid.py
```diff
@@ -2854,7 +2853,7 @@ def unsplit_x_section_points(self, axis, ref_slice0 = 0, plus_face = False, mask
             return points[:, ref_slice0, :, :]
 
     def x_section_points(self, axis, ref_slice0 = 0, plus_face = False, masked = False):
-        """Deprecated: please use unsplit_x_section_points() instead."""
+        """Deprecated: please use `unsplit_x_section_points` instead."""
 
         return self.unsplit_x_section_points(axis, ref_slice0 = ref_slice0, plus_face = plus_face, masked = masked)
```
connortann, line 2858: Given this func is deprecated, we should raise a DeprecationWarning here

PR #282, opened 2021-11-15
Integration - mesh.py - Unit tests and refactoring for surface/mesh.py

emmanesbit:


tests/test_surface.py
```diff
@@ -329,6 +324,53 @@ def test_explicit_mesh(example_model_and_crs):
     assert_array_almost_equal(persistent_mesh.full_array_ref(), mesh.full_array_ref())
 
 
+@pytest.mark.parametrize('flavour,infile,filetype', [('explicit', 'Surface_roxartext.txt', 'roxar'),
```
andy-beer, line 327: We should check that there are no IP issues with regard to including test files in these formats in the public repository.

andy-beer (approved):
This all looks fine with a quick look.

resqpy/surface/mesh.py
```diff
@@ -133,104 +133,201 @@ def __init__(self,
...
+            self._load_from_xyz_values(xyz_values)
+
+        else:
+            self._load_from_arguments(ni, nj, origin, dxyz_dij, z_values, z_supporting_mesh_uuid, crs_uuid)
 
         assert self.crs_uuid is not None
         if not self.title:
             self.title = 'mesh'
 
+    def _load_from_arguments(self, ni, nj, origin, dxyz_dij, z_values, z_supporting_mesh_uuid, crs_uuid):
```
cflynn3, line 148: Please use double underscore for private methods / properties.

cflynn3 (approved):
Aside from the naming issue mentioned this looks like a great set of changes.

PR #284, opened 2021-11-15
Refactoring organize

NirjhorC:


cflynn3 (approved):
This looks really good so far, and can be merged once you re-run the formatter, which (I think) is causing it to fail the static analysis checks.

PR #285, opened 2021-11-16
Integration - property.py - Initial integration and unit tests for property.py

emmanesbit:
Initial tests to cover most of the functionality prior to any refactoring.

### DIFF
--- a/resqpy/property.py
+++ b/resqpy/property.py
@@ -5989,7 +5989,7 @@ def selective_version_of_collection(
     if support_uuid is None and grid is not None:
         support_uuid = grid.uuid
     if support_uuid is not None:
-        view.set_support(support_uuid = support_uuid)
+        view.set_support(support_uuid = support_uuid, model = collection.model)
     if realization is not None:
         view.set_realization(realization)
     view.inherit_parts_selectively_from_other_collection(collection,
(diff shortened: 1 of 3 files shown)


emmanesbit:
@cflynn3 @andy-beer - This PR adds some extra tests for property.py, and a few very minor bug fixes. I'll add some additional tests after refactoring property.py into separate files

resqpy/property.py
```diff
@@ -2724,8 +2725,8 @@ def add_cached_array_to_imported_list(self,
             if not discrete and np.all(np.isnan(zorro)):
                 min_value = max_value = None
             else:
-                min_value = np.nanmin(zorro)
-                max_value = np.nanmax(zorro)
+                min_value = int(np.nanmin(zorro))
```
andy-beer, line 2728: The logic is not quite right here.  The int() must only be applied when discrete is True.

andy-beer (commented):
The tests look excellent.  I think there is still something wrong with the min_value, max_value code.

emmanesbit:
@andy-beer good spot on the min/max logic - this should now be resolved (and perhaps suggests another test is needed in the next PR to check min/max for non-integer continuous properties!)

resqpy/property.py
andy-beer: I think that will do it, so I'll mark this as resolved.  Could I suggest a test with a discrete array, with a null value defined (typically -1) and all the values in the array set to the null value.  Ideally the min and max values would end up as None in that scenario.  If it comes out different, we might need to strip out the int(np.nanmin()) and use a simple np.min() on the masked zorro array.

andy-beer (approved):

cflynn3 (approved):

Issue #262, opened 2021-11-09
RESQPY Refactoring - derived_model.py


cflynn3:
Refactor derived_model.py to make it easier to work with and increase its unit test coverage. Carry out the following steps:

- [ ] Break down the existing functions into smaller, less complex methods.
- [ ] Add unit tests to test each function that has been created, including as many different code paths as possible.

PR #287, opened 2021-11-16
Refactor derived model

andy-beer:
derived_model.py replaced with a package of small files
yet to do:

- reduce complexity of some of the functions
- add unit tests for functions currently untested

Resolves issue #262


resqpy/derived_model/__init__.py
```diff
@@ -0,0 +1,29 @@
...
+from .dm_extract_box import extract_box
+from .dm_fault_throw_scaling import fault_throw_scaling, global_fault_throw_scaling
+from .dm_gather_ensemble import gather_ensemble
+from .dm_interpolated_grid import interpolated_grid
+from .dm_local_depth_adjustment import local_depth_adjustment
+from .dm_refined_grid import refined_grid
+from .dm_tilted_grid import tilted_grid
+from .dm_unsplit_grid import unsplit_grid
+from .dm_zonal_grid import zonal_grid, single_layer_grid
+from .dm_zone_layer_ranges_from_array import zone_layer_ranges_from_array
```
emmanesbit, line 29: We don't seem to be importing dm_common.py - does this only contain functions used by others in this directory, not something users might call?

andy-beer: That's right, dm_common only holds 'private' functions used by some of the dm_* modules but not to be made available to calling code.

resqpy/derived_model/dm_zonal_grid.py
```diff
@@ -0,0 +1,337 @@
...
+    model.h5_release()
+    if new_epc_file:
+        __write_grid(new_epc_file, grid, grid_title = new_grid_title, mode = 'w')
+    else:
+        __write_grid(epc_file, grid, ext_uuid = None, grid_title = new_grid_title, mode = 'a')
+
+    return grid
+
+
+def single_layer_grid(epc_file,
```
emmanesbit, line 123: Where is this used?

emmanesbit: Ah I see

emmanesbit (approved):
Looks good to me

PR #291, opened 2021-11-21
Refactoring well

kadijahassanali:
Separated resqpy.well into into individual files, each containing a single class. These include:

1.  MdDatum
2.  DeviationSurvey
3.  Trajectory
4.  BlockedWell
5.  WellboreFrame
6.  WellboreMarkerFrame

 I also simplified class methods with CC > 10.

well_utils contains functions that are common to several of the new well classes
well_object_funcs contains functions that require importing the new well classes

Final CC scores:
image


resqpy/well/__init__.py
```diff
@@ -0,0 +1,18 @@
+"""resqpy modules providing trajectory, deviation survey, blocked well, wellbore frame, wellbore marker frame and
+ md datum classes.
```
connortann, line 2: Minor issue - first line needs to be a one-line, optionally follow by blank line and then any further lines.

A docstring like this that is almost a one-liner but is too long and run into two lines is unfortunately a common cause of ugly-looking HTML pages in the docs!

resqpy/well/__init__.py
```diff
@@ -0,0 +1,18 @@
...
+"""
+
+__all__ = [
+    'md_datum', 'blocked_well', 'well_object_funcs', 'trajectory', 'deviation_survey', 'wellbore_frame',
+    'wellbore_marker_frame'
```
connortann, line 7: This list should be the actual public classes and methods, not the inner submodules

resqpy/well/blocked_well.py
```diff
@@ -0,0 +1,3066 @@
+"""blocked_well.py: resqpy well module providing blocked well class.
```
connortann, line 1: I would suggest making each of these "inner" submodules private (i.e. with a single underscore prefix), to avoid exposing them as a public api. That is, `_blocked_well.py` rather than `blocked_well.py`.

This will maintain the existing api: `from resqpy.well import BlockedWell` , but not expose the alternative `from resqpy.well._blocked_well import BlockedWell`. 

In turn, this keeps us free to refactor and reshuffle things internally without it being a breaking change.

tests/test_model.py
```diff
@@ -9,6 +9,7 @@
 import resqpy.olio.write_hdf5 as rwh5
 import resqpy.olio.xml_et as rqet
 import resqpy.property as rqp
+
```
andy-beer, line 12: Not sure where this blank line has come from! (Also in test_s_bend.py)

andy-beer (commented):
Nothing further to add except to agree with Connor's suggestions.  If you've not already done so, please check that all public methods have a docstring.

cflynn3 (approved):

resqpy/well/__init__.py
kadijahassanali: image

resqpy/well/blocked_well.py
kadijahassanali: Thanks for the feedback! I've made the submodules private and fixed the init file. @andy-beer, I was wondering if I should makde the functions in resqpy\well\well_utils public also. These functions are only used by the well classes, and some have a "meta private" note in their docstring. Currently, these aren't listed in the init file.

andy-beer:
The recent merge into integration from betterdocs included the following two docstring insertions in the old well.py. Please replicate in refactoring_well branch before merge into integration:

@@ -1005,6 +1005,7 @@ def load_from_cell_list(self, grid, cell_kji0_list, spline_mode = 'cube', md_uom
         self.set_measured_depths()

     def load_from_wellspec(self, grid, wellspec_file, well_name, spline_mode = 'cube', md_uom = 'm'):
         **"""Sets the trajectory data based on visiting the cells identified in a Nexus wellspec keyword."""**

         col_list = ['IW', 'JW', 'L']
         wellspec_dict = wsk.load_wellspecs(wellspec_file, well = well_name, column_list = col_list)
 @@ -4077,6 +4078,7 @@ def create_xml(self,
                    wellbore_marker_list = None,
                    title = 'wellbore marker framework',
                    originator = None):
         **"""Creates the xml tree for this wellbore marker frame and optionally adds as a part to the model."""**

         assert type(add_as_part) is bool

kadijahassanali:
@cflynn3  Merge me please!
image

PR #295, opened 2021-11-24
Restructure unstructured

andy-beer:
unstructured.py split into resqpy class specific modules


resqpy/unstructured/__init__.py
```diff
@@ -0,0 +1,11 @@
+"""Package containing unstructured grid classes."""
+
+__all__ = [
+    'UnstructuredGrid', 'HexaGrid', 'PrismGrid', 'VerticalPrismGrid', 'PyramidGrid', 'TetraGrid', 'valid_cell_shapes'
+]
+
+from .unstructured_grid import UnstructuredGrid, valid_cell_shapes
```
connortann, line 7: May I suggest making these inner submodules "private", as per our previous Teams chat?

cflynn3 (approved):

PR #309, opened 2021-11-25
Refactoring well tests -- draft

kadijahassanali:
Adding unit and integration tests to the well classes.


resqpy/well/_wellbore_marker_frame.py
```diff
@@ -362,11 +466,18 @@ def find_marker_from_interp(self, interpetation_obj = None, uuid = None):
...
 
     def find_marker_from_index(self, idx):
-        """Returns wellbore marker by index."""
+        """Returns wellbore marker by index.
+
+        arguments:
+           idx (int): position of the marker in the wellbore marker list
+
+        returns:
+           tuple; tuple is (marker UUID, geologic boundary, marker citation title, interp. o
```
andy-beer, line 480: I think _geologic boundary_ would be better as _geologic boundary type_

resqpy/well/well_object_funcs.py
```diff
@@ -408,6 +408,8 @@ def add_blocked_wells_from_wellspec(model, grid, wellspec_file):
 
     log.info(f'{count} blocked wells created based on wellspec file: {wellspec_file}')
 
+    return count  # TODO: Kadija, confirm that this return statement should be here. You added this!
```
andy-beer, line 411: The code is consistent with the docstring, and will be a harmless change from no value being returned, so I think the TODO can be removed.

resqpy/well/well_object_funcs.py
```diff
@@ -533,7 +535,10 @@ def lookup_from_cellio(line, model):
 
     # Check if a StringLookupTable already exists in the model, with the same name and values
     for existing in model.parts_list_of_type('obj_StringTableLookup'):
-        table = rqp.StringLookup(parent_model = model, root_node = model.root_for_part(existing))
+        table = rqp.StringLookup(
+            parent_model = model,
+            uuid = model.uuid_for_part(part_name = existing))  #TODO: confirm it is okay to change root_node to uuid
```
andy-beer, line 540: Confirmed that use of uuid is much preferable to root_node.

I would recommend changing the for statement to:
`for existing_uuid in model.uuids(obj_type = 'StringTableLookup'):`

resqpy/well/_wellbore_frame.py
```diff
@@ -161,7 +161,8 @@ def create_feature_and_interpretation(self):
 
         Uses the wellboreframe citation title as the well name
         """
-        if self.wellbore_interpretation is not None:
+        # if self.wellbore_interpretation is not None: # Kadija: I think this may have been an error
```
andy-beer, line 164: Yes it looks like the old code was in error.

resqpy/well/_trajectory.py
```diff
@@ -997,3 +997,5 @@ def __eq__(self, other):
         # TODO: more detailed equality comparison
         other_uuid = getattr(other, "uuid", None)
         return isinstance(other, self.__class__) and bu.matching_uuids(self.uuid, other_uuid)
+        other_uuid = getattr(other, "uuid", None)
```
andy-beer, line 1000: These two lines appear to be duplicates of the two above, and will never execute due to the earlier return statement.

tests/test_deviation_survey.py
```diff
@@ -0,0 +1,141 @@
...
+    # Load example model from a fixture
+    model, well_interp, datum, traj = example_model_with_well
+    epc_path = model.epc_file
+
+    # Create 3 copies of the survey, using different initialisers
+    data = dict(
+        title = 'Majestic Umlaut ö',
+        originator = 'Thor, god of sparkles',
+        md_uom = 'ft',
+        angle_uom = 'rad',
```
andy-beer, line 26: The inclinations need to be between +/- 180 degrees, so between +/- π radians.

Azimuths need to be between 0 and 360 degrees, or 0 and 2π radians (which the current data is)

tests/test_trajectory.py
```diff
@@ -0,0 +1,305 @@
...
+    mds = np.array([300, 310, 330, 340])
+    zs = mds - elevation
+    source_dataframe = pd.DataFrame({
+        'MD': mds,
+        'X': [1, 2, 3, 4],
+        'Y': [1, 2, 3, 4],
+        'Z': zs,
+        'WELL': ['Coconut', 'Coconut', 'Mango', 'Mango']
+    })
+    deviation_survey_file_path = os.path.join(model.epc_directory, 'deviation_survey.csv')
```
andy-beer, line 138: strictly speaking this is trajectory data rather than a deviation survey (which measures inclinations and azimuths); in the industry the derived trajectory is often lazily referred to as a deviation survey but as RESQML makes a clear distinction, it might help if we stick with RESQML compatible variable names

tests/test_trajectory.py
```diff
@@ -0,0 +1,305 @@
...
+    # --------- Assert ----------
+    assert trajectory_from_wellspec is not None
+    assert trajectory_from_wellspec.knot_count == len(trajectory_from_wellspec.control_points)
+
+
+def test_load_from_ascii_file(example_model_and_crs):
+
+    # --------- Arrange ----------
+    model, crs = example_model_and_crs
+    well_names = [None, 'Coconut']
```
andy-beer, line 122: What? No Mango? What's the reason? Is it the wrong season?

tests/test_trajectory.py
```diff
@@ -0,0 +1,305 @@
...
+                                        length_uom = 'm')
+
+    # --------- Act ----------
+    # Calculate tangent vectors based on control points
+    trajectory.set_tangents()
+
+    # --------- Assert ----------
+    assert trajectory.tangent_vectors is not None
+    assert trajectory.tangent_vectors.shape == (trajectory.knot_count, 3)
+
```
andy-beer, line 200: You could use some vector_utilities to perform some further basic checks on the tangents, eg.:
naive length of each tangent is close to 1
with the xyz data you are using, I would expect tangent[1] and tangent[2] to be more-or-less the same

tests/test_trajectory.py
```diff
@@ -0,0 +1,305 @@
...
+                                        data_frame = source_dataframe,
+                                        well_name = well_name,
+                                        md_datum = datum,
+                                        length_uom = 'm')
+    # --------- Act ----------
+    # # Get the xyz triplet for a given measured depth
+    x, y, z = trajectory.xyz_for_md(md = 305)
+
+    # # -------- Assert ---------
+    assert x == y == 1.5
```
andy-beer, line 234: better to use maths.isclose() rather than an exact == when testing computed floating point values
(where maths is what I import math as)

tests/test_trajectory.py
```diff
@@ -0,0 +1,305 @@
...
+
+    # Create a trajectory from dataframe
+    trajectory = resqpy.well.Trajectory(parent_model = model,
+                                        data_frame = source_dataframe,
+                                        well_name = well_name,
+                                        md_datum = datum,
+                                        length_uom = 'm')
+    # --------- Act ----------
+    # Get splined trajectory
+    splined_trajectory = trajectory.splined_trajectory(well_name = 'Coconut')
```
andy-beer, line 267: As your basic trajectory data is slightly impossible, I'm not totally sure of the behaviour of the splined_trajectory() method in these conditions.

tests/test_well_object_funcs.py
```diff
@@ -0,0 +1,208 @@
...
+            row = source_df.iloc[row_index]
+            for col in source_df.columns:
+                if col == 'WELL':
+                    fp.write(f' {(row[col]):6s}')
+                else:
+                    fp.write(f' {row[col]:6.2f}')
+            fp.write('\n')
+
+    well_col_name = 'Wellname'
+    with pytest.raises(AssertionError) as exc:
```
andy-beer, line 35: Does the testing include a successful execution of the add_wells_from_ascii_file() function, as well as this failure mode test.

tests/test_wellbore_marker_frame.py
```diff
@@ -0,0 +1,433 @@
...
+                                                 boundary_relation_list = ['conformable'])
+    horizon_interp_1.create_xml()
+
+    horizon_feature_2 = rqo.GeneticBoundaryFeature(parent_model = model,
+                                                   kind = 'horizon',
+                                                   feature_name = 'horizon_feature_2')
+    horizon_feature_2.create_xml()
+    horizon_interp_2 = rqo.HorizonInterpretation(parent_model = model,
+                                                 title = 'horizon_interp_2',
+                                                 genetic_boundary_feature = horizon_feature_1,
```
andy-beer, line 227: Did you intend this to be set to horizon_feature_1?  or 2?

tests/test_wellbore_marker_frame.py
```diff
@@ -0,0 +1,433 @@
...
+        'Well': well_name
+    })
+
+    # Create a wellbore marker frame from a dataframe
+    wellbore_marker_frame = resqpy.well.WellboreMarkerFrame.from_dataframe(parent_model = model,
+                                                                           dataframe = wellbore_marker_frame_dataframe,
+                                                                           trajectory = trajectory)
+    # --------- Act ----------
+    # Get an interpretation based on an uuid
+    found_interpretation = wellbore_marker_frame.get_interpretation_obj(interpretation_uuid = horizon_interp_2_uuid)
```
andy-beer, line 258: This is fine for now as a test. Could you please add to the comment to say that this method will have breaking changes (or be removed) sometime soon!

tests/test_wellbore_marker_frame.py
```diff
@@ -0,0 +1,433 @@
...
+    })
+
+    # Create a wellbore marker frame from a dataframe
+    wellbore_marker_frame = resqpy.well.WellboreMarkerFrame.from_dataframe(parent_model = model,
+                                                                           dataframe = wellbore_marker_frame_dataframe,
+                                                                           trajectory = trajectory)
+    print(wellbore_marker_frame.wellbore_marker_list)
+    # --------- Act ----------
+    # Get the marker object from an interpretation object and interpretation object uuid
+    found_marker1 = wellbore_marker_frame.find_marker_from_index(1)
```
andy-beer, line 430: If I recall, this method is due to be changed from 1 based indexing to pythonic zero based – which shouldn't affect the result of this test as it is written. However, it might be better to make the assertion(s) specific to the current implementation of the method, so that it will fail when we change the definition of the method. That way, we will be more sure that we have made the change as we intended. (Of course we will have to change the test to get it to pass again, once we've seen it fail.)

tests/test_blocked_well.py
```diff
@@ -0,0 +1,588 @@
...
+    well_name = 'DOGLEG'
+    bw = resqpy.well.BlockedWell(model,
+                                 well_name = well_name,
+                                 use_face_centres = True,
+                                 grid = grid,
+                                 add_wellspec_properties = True)
+    col_list_orig = ['IW', 'JW', 'L']
+
+    # --------- Act ----------
+    result = bw._BlockedWell__derive_from_wellspec_check_grid_name(check_grid_name = check_grid_name,
```
andy-beer, line 101: Is this the syntax for calling @staticmethods? Maybe functions like this shouldn't be part of the class at all. I would be inclined to take it out of the BlockedWell class and stick it at the end of the module as a normal standalone function. But others might have other opinions.

tests/test_blocked_well.py
```diff
@@ -0,0 +1,588 @@
...
+
+    # --------- Act ----------
+    # populate empty blocked well object for a 'vertical' well in the given column
+    bw.derive_from_cell_list(cell_kji0_list = cell_kji0_list, well_name = well_name, grid = grid)
+    # get list of grid uuids associated with this blocked well
+    grid_uuid_list = bw.grid_uuid_list()
+
+    # --------- Assert ----------
+    assert len(grid_uuid_list) == 1
+    assert grid_uuid_list[0] == grid_uuid
```
andy-beer, line 186: In general it is better to compare uuids using bu.matching_uuids() – where bu is what I tend to import resqpy.olio.uuid as. In general in resqpy, uuids can be passed around either as a UUID object or as a hexadecimal string (in the approved ISO format with hyphens in all the right places). The matching_uuids() allows each of the two arguments to be in either form and still returns the desired comparison.

tests/test_blocked_well.py
```diff
@@ -0,0 +1,588 @@
...
+    cell_kji0_list = np.array([(1, 1, 1), (2, 2, 2), (3, 3, 3), (3, 3, 4)])
+    bw = resqpy.well.BlockedWell(model, well_name = well_name, use_face_centres = True, add_wellspec_properties = True)
+
+    # --------- Act ----------
+    # populate empty blocked well object for a 'vertical' well in the given column
+    bw.derive_from_cell_list(cell_kji0_list = cell_kji0_list, well_name = well_name, grid = grid)
+
+    # --------- Assert ----------
+    # no tail added to trajectory measured depths as well terminates in 4th layer
+    assert bw.node_count == len(bw.node_mds) == len(bw.trajectory.measured_depths)
```
andy-beer, line 158: Please add an assertion that the blocked well cell count matches the length of the cell_kji0_list that was used to set it up.

andy-beer (commented):
Overall this is good and I'm happy for the changes to go into the integration branch, perhaps after having a look at the comments I've made.  One more general comment is that many of the tests check that some values have been set without checking that the values are correct.  But that is fair enough as to test for correct values would often mean writing the whole functionality of the methods being tested all over again, in the test code, which is not practical.

kadijahassanali:
Really appreciate the feeback @andy-beer. I'll make all your recommended changes today. 

Your comment on testing the correctness of the values is a fair one. Are there any particular methods that you think I should absolutely replicate the calculations to ensure the correct value is produced?

tests/test_trajectory.py
kadijahassanali: Sadly there are no mangoes on my Julie mango tree right now #tragictimes

tests/test_trajectory.py
kadijahassanali: I didn't appreciate the difference before you pointed it out! As RESQML does make a clear distinction, I've added a TODO statement to change the argument name in the Trajectory_init.

tests/test_blocked_well.py
cflynn3: Agree with @andy-beer here. I think all of the derive_from_wellspec... methods could be converted to functions and put into their own file.

tests/test_trajectory.py
kadijahassanali: Considering I used the most basic, x = y line for the X and Y coordinates, it should have been obvious that the azimuths of the tangent vectors would all be 45 :D.

tests/test_trajectory.py
kadijahassanali: My thought process was that the splined trajectory end points should be almost equal to the original trajectory end points so that's why I added those assertions. Was that assumption completely off-base?

tests/test_well_object_funcs.py
kadijahassanali: now it does :D

tests/test_wellbore_marker_frame.py
kadijahassanali: I changed the assertion statement to:
assert found_marker1[3] == wellbore_marker_frame['Type'][2]. The test will fail once we change the indexing in the method to zero-based

PR #313, opened 2021-11-26
Fix docs for nested subpackages

### LINKED ISSUE #310: Refactored subpackages not appearing in HTML docs
I triggered a build of the docs for the `integration` branch, and noticed that classes and methods corresponding to refactored sub-packages are completely missing e.g. `resqpy.lines` :

image

connortann:
Closes #310

Discussion of the fix in the linked issue above

### DIFF
--- a/resqpy/lines/_polyline.py
+++ b/resqpy/lines/_polyline.py
@@ -548,12 +548,13 @@ class Polyline(_BasePolyline):
         if rep_int_root is None:
             rep_int_root = self.rep_int_root  # todo: check whether it is legal to have 2 representations for 1 interpretation
 
-        return Polyline(self.model,
-                        set_bool = self.isclosed,
-                        set_coord = spline_coords,
-                        set_crs = self.crs_uuid,
-                        title = title,
-                        rep_int_root = rep_int_root)
+        cls = type(self)  # Get class constructor object i.e. Polyline
+        return cls(self.model,
+                   set_bool = self.isclosed,
+                   set_coord = spline_coords,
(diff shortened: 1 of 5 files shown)


resqpy/lines/_polyline.py
```diff
@@ -548,12 +548,13 @@ def splined(self,
         if rep_int_root is None:
             rep_int_root = self.rep_int_root  # todo: check whether it is legal to have 2 representations for 1 interpretation
 
-        return Polyline(self.model,
-                        set_bool = self.isclosed,
-                        set_coord = spline_coords,
-                        set_crs = self.crs_uuid,
-                        title = title,
-                        rep_int_root = rep_int_root)
+        cls = type(self)  # Get class constructor object i.e. Polyline
```
andy-beer, line 551: This is an example of a code change which I regard as logical but if anything making the code harder to read. If the class name ever changes in future, it would be a significant breaking change requiring a sweep of the code base anyway, so having one less change to make would not save much in that unlikely event. Having said all that, I don't feel strongly enough about it to request the change is undone.

andy-beer (approved):

resqpy/lines/_polyline.py
connortann: Ah yes. I changed that whilst debugging, as I was experimenting changing the class name to `_Polyline` and the in the init using `import _Polyline as Polyline`, and saw that change broke the code. But then I renamed the class back how it was, so the change above isn't needed. I'll put it back, agree with you the original was clearer

cflynn3 (approved):

Issue #323, opened 2021-11-30
xml escape characters not being written correctly


andy-beer:
xml requires use of an escape sequence when special characters & < > appear in a string, and in some cases for " '

At present the resqpy code does not write the required escape sequence, resulting in invalid xml

PR #324, opened 2021-11-30
xml escape character bug fix; resolves issue #323

andy-beer:


andy-beer:
This pull request seems to have incorporated lots of unintended changes from elsewhere, so I'm closing it and will try again.

PR #325, opened 2021-11-30
Refactor - Property collection refactor

emmanesbit:


resqpy/property/_property_collection_get_attributes.py
```diff
@@ -0,0 +1,719 @@
...
+
+def _realizations_array_ref_get_r_extent(fill_missing, r_list):
+    if fill_missing:
+        r_extent = r_list[-1] + 1
+    else:
+        r_extent = len(r_list)
+    return r_extent
+
+
+def _add_part_to_dict_get_count_and_indexable(xml_node):
```
andy-beer, line 460: This and the other _add_part_* functions might be better off living in _property_collection_add_part.py as they have to do with extracting data from xml during the adding of a property to the collection. This module would be better restricted to the functions that are inspecting the collection once it is constructed.

resqpy/property/property_collection.py
```diff
@@ -15,13 +15,14 @@
...
 import resqpy.olio.xml_et as rqet
-import resqpy.time_series as rts
 from resqpy.olio.xml_namespaces import curly_namespace as ns
 
-from .property_kind import PropertyKind
 from .string_lookup import StringLookup
-from .property_common import same_property_kind, supported_property_kind_list, property_kind_and_facet_from_keyword, \
-    dtype_flavour, _cache_name, _cache_name_for_uuid, selective_version_of_collection, guess_uom
+from .property_common import dtype_flavour, _cache_name, _cache_name_for_uuid, selective_version_of_collection
+import resqpy.property._property_collection_create_xml_functions as pcxml
```
andy-beer, line 22: Personally I have yoyo-ed between including the package identification in each module name, or not. At the moment I favour shorter file names without the package being repeated in each file name, so:
resqpy.property._create_xml_functions
instead of
resqpy.property._property_collection_create_xml_functions

But I see pro's and con's for both.

andy-beer (commented):
I've only skimmed through the changes but it looks good at a glance.  Apart from moving a few functions from ..._get_attributes.py to ..._add_part.py, I approve.

resqpy/property/property_collection.py
emmanesbit: I have trimmed the names somewhat, though leaving in the reference to 'collection' for now. Once the other property classes have been refactors I may revisit and remove the collection reference if possible

andy-beer (approved):

cflynn3 (approved):

PR #327, opened 2021-12-01
Additional integration tests for property.py

emmanesbit:
This PR adds some additional tests for property.py. These new tests have been written and run against the master branch, to ensure no breaking changes due to refactoring.


emmanesbit:
@cflynn3 @andy-beer - this one is ready for review

tests/test_property.py
```diff
@@ -508,14 +509,14 @@ def test_remove_part_from_dict(example_model_with_properties):
     model = example_model_with_properties
     pc = model.grid().property_collection
     assert pc is not None
-    assert len(pc.parts()) == 7
+    assert len(pc.parts()) == 8
```
cflynn3, line 512: Is this an expected change in behaviour in the code?

cflynn3 (approved):

tests/test_property.py
andy-beer: 8 is the correct expectation here: the example_model_with_properties builds 7 explicit grid properties but the grid included in that model has the (default) addition of an active cell property. I don't think the resqpy code has changed in that regard for a very long time.

tests/conftest.py
```diff
@@ -280,7 +286,7 @@ def example_model_with_prop_ts_rels(tmp_path):
     # Add non-varying properties
     for array, name, kind, discrete, facet_type, facet in zip([zone_array, vpc_array, fb_array, perm_array],
                                                               ['Zone', 'VPC', 'Fault block', 'Perm'],
-                                                              ['discrete', 'discrete', 'discrete', 'rock permeability'],
+                                                              ['discrete', 'discrete', 'discrete', 'permeability rock'],
```
andy-beer, line 289: in the RESQML units of measure there is one instance of a synonym, and this is it: 'rock permeability' and 'permeability rock' should be interchangeable. We might want to modify the resqpy code, which I think was generating an erroneous warning, to handle this special situation.

tests/test_property.py
```diff
@@ -874,3 +875,479 @@ def test_set_support_mesh(example_model_and_crs):
 
     pc = rqp.PropertyCollection()
     pc.set_support(support = support)
```
andy-beer, line 877: Could be worth adding an assertion that the support uiud associated with pc does indeed match support.uuid

tests/conftest.py
emmanesbit: Ah I didn't realise permeability rock was legal, and assumed a typo on my side. I will update the code to allow for the opposite, and add an additional test to ensure this is allowed

tests/test_property.py
emmanesbit: Updated in latest commit

tests/conftest.py
emmanesbit: Issue added for future update

andy-beer (approved):
That's an impressive bundle of tests for the basic static properties method, as well as for the normalisation!

andy-beer:
Looks like a yapf is needed on __init__.py

PR #331, opened 2021-12-06
Refactoring - ongoing test addition and refactoring for property_common.py

emmanesbit:


emmanesbit:
@andy-beer @cflynn3 - this PR is ready for review. This contains additional tests for property_common.py, mostly for property_kind_and_facet_for_keyword and guess_uom, and refactoring of these two functions.

resqpy/property/property_common.py
```diff
@@ -307,111 +367,129 @@ def guess_uom(property_kind, minimum, maximum, support, facet_type = None, facet
...
+    if 'thickness' in property_kind:
         z_units = rqet.find_tag(crs_node, 'VerticalUom').text.lower()
         if z_units == 'm':
             return 'mD.m'
         if z_units == 'ft':
             return 'mD.ft'
         return None
-    if property_kind == 'permeability length':
+    else:
         xy_units = rqet.find_tag(crs_node, 'ProjectedUom').text.lower()
```
andy-beer, line 436: This should use the z units instead of the xy units for the length component, I believe.

andy-beer (changes requested):
This all looks good.  Could you please make the one small change commented upon, then good to merge.

We should also do some careful integration testing using rq_import with vdb input when we get to that stage.

resqpy/property/property_common.py
emmanesbit: Updated to return None if the xy / z units are mismatched, or the xyz units if matching.

andy-beer (approved):

PR #338, opened 2021-12-08
Refactoring and unit tests for time series code

andrewedirisooriya:
Refactoring and unit tests for time series code


requirements.txt
```diff
@@ -0,0 +1,7 @@
+pytest~=6.2.5
```
andy-beer, line 1: resqpy dependencies should be handled via existing configuration files – what is this file for?

resqpy/time_series/__init__.py
```diff
@@ -0,0 +1,13 @@
+"""Time series module"""
+
+__all__ = [
+    'time_series', 'time_series_from_nexus_summary', 'time_duration', 'any_time_series_base_resqpy', 'TimeSeries'
```
andy-beer, line 4: This list should contain the previously visible classes and functions without any name changes (unless we need to make a breaking change for a good reason).  By my reckoning: ['AnyTimeSeries', 'GeologicTimeSeries', 'TimeDuration', 'TimeSeries', 'selected_time_series', 'simplified_timestamp', 'cleaned_timestamp', 'time_series_from_list', 'merge_timeseries_from_uuid', 'time_series_from_nexus_summary', 'geologic_time_str', 'timeframe_for_time_series_uuid', 'any_time_series']

And then each of those must be imported from its new home in the from... statements below.

resqpy/time_series/any_time_series_base_resqpy.py
```diff
@@ -0,0 +1,212 @@
+"""Any time series base resqpy"""
```
andy-beer, line 1: I would rather the module name did not include the inherited class, so any_time_series.py

Even better, like Connor I prefer all the submodules to be flagged as private by giving them file names starting with _, so _any_time_series.py

resqpy/time_series/geologic_time_series_any_time_series.py
```diff
@@ -0,0 +1,84 @@
+"""Geologic time series"""
```
andy-beer, line 1: Comments as for  any_time_series_base_resqpy.py

resqpy/time_series/time_duration.py
```diff
@@ -0,0 +1,63 @@
+"""Time duration"""
+
+import logging
+
+log = logging.getLogger(__name__)
+log.debug('resqml_time_series.py version ')
```
andy-beer, line 6: Please remove all these log.debug(...version..) statements.

resqpy/time_series/time_series.py
```diff
@@ -0,0 +1,158 @@
+"""time_series.py: RESQML time series class."""
```
andy-beer, line 1: I would rather this module was called _time_series_functions.py or _time_series_fns.py or just _functions.py if you prefer.

resqpy/time_series/time_series.py
```diff
@@ -0,0 +1,158 @@
...
+
+version = '16th September 2021'
+
+# Nexus is a registered trademark of the Halliburton Company
```
andy-beer, line 5: This comment should be included in files, and only those files, where 'Nexus' appears somewhere.

resqpy/time_series/time_series_any_time_series.py
```diff
@@ -0,0 +1,233 @@
+"""Any time series"""
```
andy-beer, line 1: As with the other class modules, I'd prefer to drop the inherited class from the filename, so: _time_series.py (which clearly requires the module containing functions to be named something else!)

resqpy/time_series/time_series_from_nexus_summary.py
```diff
@@ -0,0 +1,92 @@
...
+        if entry[1] < last_cumulative_time:  # should never happen
+            time_series.extend_by_days(0)
+        else:
+            time_series.extend_by_days(entry[1] - last_cumulative_time)
+            last_cumulative_time = entry[1]
+        last_timestep += 1
+    return time_series
+
+
+def open_file(summary_file):
```
andy-beer, line 43: This function should be private to this module, so _open_file()

resqpy/time_series/time_series_from_nexus_summary.py
```diff
@@ -0,0 +1,92 @@
...
+
+log = logging.getLogger(__name__)
+log.debug('resqml_time_series.py version ')
+
+import datetime as dt
+import os
+from .time_series_any_time_series import TimeSeries
+
+
+def process_summary_entries(summary_entries, parent_model = None):
```
andy-beer, line 13: This function should be private to this module, so _process_summary_entries()

resqpy/time_series/time_series_from_nexus_summary.py
```diff
@@ -0,0 +1,92 @@
...
+                date = dt.date(int(date_str[-4:]), int(date_str[:-8]), int(date_str[-7:-5]))
+            else:
+                date = dt.date(int(date_str[-4:]), int(date_str[-7:-5]), int(date_str[:-8]))
+            summary_entries.append((ts_number, time_in_days, date))
+    return summary_entries
+
+
+def time_series_from_nexus_summary(summary_file, parent_model = None):
+    """Create a TimeSeries object based on time steps reported in a nexus summary file (.sum)."""
+
```
andy-beer, line 78: Please add a note to this docstring:

note:
   this function does not create the xml for the new TimeSeries, nor add it as a part to the parent model.

tests/test_timeseries.py
```diff
@@ -118,3 +115,21 @@ def test_geologic_time_series(tmp_path):
...
+
+def test_geologic_time_str_fails_when_not_int():
+    with pytest.raises(AssertionError):
+        rqts.geologic_time_str('hello world')
+
+
+@pytest.mark.parametrize('years_value, expected_result', [(10_000_000, '-10 Ma'), (1_000_000, '-1 Ma'),
+                                                          (11_000_000, '-11 Ma'), (1, '-0.000 Ma'),
+                                                          (1_500_000, '-1.500 Ma'), (15_060_000, '-16 Ma'),
+                                                          (15_040_000, '-16 Ma')])
```
andy-beer, line 128: Interesting...we might want to change the geologic_time_str() function to round() to the nearest Ma, rather than round down.  But we can park that for a separate change some other time.

tests/unit_tests/test_time_duration.py
```diff
@@ -0,0 +1,29 @@
...
+import pytest
+import datetime as dt
+
+log = logging.getLogger(__name__)
+log.debug('resqml_time_series.py version ')
+
+
+@pytest.mark.parametrize('later_timestamp, later_timestamp_end',
+                         [('2014-01-01T00:00:00.588Z', '2014-01-01T00:00:00.588')])
+def test_timestamp_before_duration(later_timestamp, later_timestamp_end):
```
andy-beer, line 11: This function is not testing resqpy code.

tests/unit_tests/test_time_duration.py
```diff
@@ -0,0 +1,29 @@
...
+    test_later_timestamp = later_timestamp
+    test_later_timestamp_end = later_timestamp_end
+    # act
+    test_later_timestamp = test_later_timestamp[:-1]
+    # assert
+    assert test_later_timestamp == test_later_timestamp_end
+
+
+@pytest.mark.parametrize('later_timestamp_end', ['2014-01-01T00:00:00.588'])
+def test_timestamp_iso(later_timestamp_end):
```
andy-beer, line 22: This function is also not testing any resqpy code as it stands.

tests/unit_tests/test_time_series_any_time_series.py
```diff
@@ -0,0 +1,293 @@
...
+
+
+@pytest.mark.parametrize('timestamps', [['2014-01-01T00:00:00.588Z', '2015-02-02T00:00:00.588Z']])
+def test_index_for_timestamp_closest_to_empty_timestamps(timestamps, example_model_and_crs):
+    # arrange
+    model, crs = example_model_and_crs
+    any_time_series = TimeSeries(model)
+    any_time_series.timestamps = None
+    # act
+    result = any_time_series.index_for_timestamp_closest_to('2000-01-01T00:00:00.588Z')
```
andy-beer, line 90: Do we expect this to return None when passed a valid timestamp outside the range of the TimeSeries timestamps?  If so, I feel the docstring for the function should make that clear.  Otherwise I would expect this call to return an index of 0.

tests/unit_tests/test_time_series_any_time_series.py
```diff
@@ -0,0 +1,293 @@
...
+    any_time_series = TimeSeries(model)
+    any_time_series.timestamps = timestamps
+    # act
+    result = any_time_series.index_for_timestamp_closest_to('2013-02-02T00:00:00.588Z')
+    # assert
+    assert result == 0
+
+
+@pytest.mark.parametrize('timestamps', [['2014-01-01T00:00:00.588Z', '2015-02-02T00:00:00.588Z']])
+def test_index_for_timestamp_closest_to_timestamp_len(mocker, timestamps, example_model_and_crs):
```
andy-beer, line 108: I don't understand the use of a mock function here.

tests/unit_tests/test_time_series_any_time_series.py
```diff
@@ -0,0 +1,293 @@
...
+    model, crs = example_model_and_crs
+    any_time_series = TimeSeries(model)
+    any_time_series.timestamps = timestamps
+    index_for_timestamp_not_later_mock = mocker.Mock(return_value = 0)
+    any_time_series.index_for_timestamp_not_later_than = index_for_timestamp_not_later_mock
+    # act
+    result = any_time_series.index_for_timestamp_closest_to('2014-01-01T00:00:00.588Z')
+    # assert
+    assert result == 0
+    index_for_timestamp_not_later_mock.assert_called_once_with('2014-01-01T00:00:00.588Z')
```
andy-beer, line 133: I disagree with the use of mocker functions here.  Simply test the functionality of the index_for_timestamp_closest_to() method without checking how many times it has called other internal methods.

tests/unit_tests/test_time_series_any_time_series.py
```diff
@@ -0,0 +1,293 @@
...
+
+
+@pytest.mark.parametrize('timestamps', [['2014-01-01T00:00:00.588Z', '2015-02-02T00:00:00.588Z']])
+def test_index_for_timestamp_closest_to_timestamp_early_delta(mocker, timestamps, example_model_and_crs):
```
andy-beer, line 137: Also no need for mocker

tests/unit_tests/test_time_series_any_time_series.py
```diff
@@ -0,0 +1,293 @@
...
+    # act
+    result = any_time_series.index_for_timestamp_closest_to('2014-02-02T00:00:00.588Z')
+    # assert
+    assert result == 0
+    index_for_timestamp_not_later_mock.assert_called_once_with('2014-02-02T00:00:00.588Z')
+
+
+@pytest.mark.parametrize('timestamps',
+                         [['2014-01-01T00:00:00.588Z', '2015-02-02T00:00:00.588Z', '2016-03-02T00:00:00.588Z']])
+def test_index_for_timestamp_closest_to_timestamp_later_delta(mocker, timestamps, example_model_and_crs):
```
andy-beer, line 153: Also no value in using mocker here.

tests/unit_tests/test_time_series_any_time_series.py
```diff
@@ -0,0 +1,293 @@
...
+    # act
+    any_time_series.add_timestamp(new_timestamp, allow_insertion = True)
+    # assert
+    assert any_time_series.timestamps[0] == '2014-01-01T00:00:00.588Z'
+    assert any_time_series.timestamps[1] == '2014-01-03T00:00:00.588Z'
+    assert any_time_series.timestamps[2] == new_timestamp
+
+
+@pytest.mark.parametrize('timestamps', [['2014-01-01T00:00:00.588Z', '2014-01-03T00:00:00.588Z']])
+def test_add_timestamp_between(timestamps, example_model_and_crs):
```
andy-beer, line 282: Might also be good to test that an exception is raised if allow_insertion is False with this data.

tests/unit_tests/test_time_series_from_nexus.py
```diff
@@ -0,0 +1,127 @@
...
+
+def test_time_series_from_nexus_summary_if_no_file():
+    # arrange
+    # act
+    result = time_series_from_nexus_summary('')
+    # assert
+    assert result is None
+
+
+# def test_time_series_from_nexus_summary_open_file(mocker):
```
andy-beer, line 115: At some point, we will need a test for the time_series_from_nexus_summary() either using mock data or a test data file.

andy-beer (changes requested):
Despite all my pesky comments, this refactoring and unit testing is shaping up nicely.  I'm happy to discuss anything if you'd like to @andrewedirisooriya.

tests/unit_tests/test_time_series_any_time_series.py
andy-beer: Ah, I see, my mistake – the method returns None if and only if there are no timestamps in the series.

andy-beer (approved):
Ready to go from my side.

@cflynn3 , there are some odd uses of a mocker in test_time_series – could you check that please.

cflynn3 (approved):

PR #340, opened 2021-12-08
Breaking changes WellboreMarkerFrame class

kadijahassanali:
Breaking changes made to resqpy/well/WellboreMarkerFrame class, including the creation of a new WellboreMarker class.

Tests written for new version of WellboreMarkerFrame and WellboreMarker.


kadijahassanali:
Hi @andy-beer,
One of the main questions I have is around the class method to initialise a WellboreMarkerFrame object from a list of WellboreMarker objects. 
I thought we might run into a circular reference issue since each of the markers needs a parent wellbore marker frame to be specified during initialisation. As a result, I've written a from_marker_list method that's not a class method. As usual, feedback is appreciated!
Thanks!

resqpy/well/_wellbore_marker.py
```diff
@@ -0,0 +1,167 @@
...
+        wbm_node = self.model.new_obj_node('WellboreMarker', is_top_lvl_obj = False)
+        wbm_node.set('uuid', str(self.uuid))
+
+        # Citation block
+        if self.title is not None:
+            title = self.title
+        citation = self.model.create_citation(root = wbm_node, title = title, originator = self.originator)
+
+        # Add sub-elements to root node
+        boundary_feature_dict = {
```
andy-beer, line 107: I generally prefer to put constant data like this dict outside the method, either as a class variable or a global variable outside the class.

resqpy/well/_wellbore_marker.py
```diff
@@ -0,0 +1,167 @@
...
+        assert citation_tag is not None
+        self.title = rqet.find_tag_text(root = citation_tag, tag_name = 'Title')
+        self.originator = rqet.find_tag_text(root = citation_tag, tag_name = 'Originator')
+
+        for boundary_feature_type in ['GeologicBoundaryKind', 'FluidMarker', 'FluidContact']:
+            found_tag_text = rqet.find_tag_text(root = marker_node, tag_name = boundary_feature_type)
+            if found_tag_text is not None:
+                self.marker_type = found_tag_text
+                break
+
```
andy-beer, line 162: I think we should assert that self.marker_type is not None here (having set it to None before the for loop)

tests/test_trajectory.py
```diff
@@ -5,10 +5,62 @@
...
+
+    # Reload from disk
+    model2 = Model(epc_file = epc_path)
+    trajectory2 = resqpy.well.Trajectory(model2, uuid = trajectory_uuid)
+
+    # --------- Assert --------------
+    # Check all attributes were loaded from disk correctly
+    assert trajectory2 is not None
+    assert trajectory2.well_name == well_name
+    np.testing.assert_equal(trajectory2.measured_depths, mds)
```
andy-beer, line 59: I think it's better to use the assert_array_almost_equal (from numpy.testing) when comparing floating point array data.

resqpy/well/_wellbore_marker_frame.py
```diff
@@ -349,32 +294,10 @@ def create_xml(self,
...
-                elif 'FaultInterpretation' in str(type(marker[3])):
-                    self.model.create_ref_node('Interpretation',
-                                               rqet.find_tag(rqet.find_tag(interp_root, 'Citation'), 'Title').text,
-                                               bu.uuid_from_string(interp_root.attrib['uuid']),
-                                               content_type = 'obj_FaultInterpretation',
-                                               root = wbm_node_obj)
-
+        if self.marker_list is not None:  # TODO: confirm that you can have an empty wellbore marker frame with no marker list
+            for marker in self.marker_list:
+                wbm_node_obj = marker.create_xml(parent_node = wbm_node, title = 'wellbore marker')
```
andy-beer, line 299: We could consider putting a more complicated title in here, something including a well name and feature name.

resqpy/well/_wellbore_marker_frame.py
```diff
@@ -115,130 +108,97 @@ def from_dataframe(cls,
...
-        else:
-            # create new trajectory object
-            trajectory_root_node = self.model.root_for_uuid(trajectory_uuid)
-            assert trajectory_root_node is not None, 'referenced wellbore trajectory missing from model'
-            return Trajectory(self.model, uuid = trajectory_uuid)
+        return row_interp_uuid
 
-    def get_interpretation_obj(self, interpretation_uuid, interp_type = None):
-        """Creates an interpretation object; returns a horizon or fault interpretation object.
+    # TODO: discuss making this a regular method as the markers require a parent wellbore marker frame
```
andy-beer, line 180: Yes, let's talk this over again: perhaps we want to be able to instantiate WellboreMarker objects with None as the parent marker frame, then this (as a class method) would fill in that attribute.  But that might be overkill.

andy-beer: As it stands, the function doesn't really deliver much. I think we either scrap it, adopt what I wrote just above, or modify it to be class method that creates a marker frame with a subset of markers from an existing frame in some way.  Let's discuss.

tests/test_wellbore_marker_frame.py
```diff
@@ -194,8 +146,8 @@ def test_get_interpretation_object(example_model_and_crs):
     zs = mds - elevation
     source_dataframe = pd.DataFrame({
         'MD': mds,
-        'X': [1, 2, 3],
-        'Y': [1, 2, 3],
+        'X': [150, 165, 180],
```
andy-beer, line 149: Better to use floating point numbers, 150.0 etc., in case behaviour changes with integers.

tests/test_wellbore_marker_frame.py
```diff
@@ -240,35 +186,62 @@ def test_get_interpretation_object(example_model_and_crs):
...
+                                            title = 'woc_marker')
+
+    fault_marker = resqpy.well.WellboreMarker(parent_model = model,
+                                              parent_frame = wellbore_marker_frame,
+                                              marker_index = 2,
+                                              marker_type = 'fault',
+                                              interpretation_uuid = fault_interp_1_uuid,
+                                              title = 'fault_marker')
+    marker_list = [horizon_marker, woc_marker, fault_marker]
+    node_mds = np.array([400, 410, 430])
```
andy-beer, line 223: Again, floating points please.

tests/test_wellbore_marker_frame.py
```diff
@@ -240,35 +186,62 @@ def test_get_interpretation_object(example_model_and_crs):
...
-    found_interpretation = wellbore_marker_frame.get_interpretation_obj(interpretation_uuid = horizon_interp_2_uuid)
+    # Add attributes to the wellbore marker frame from the marker list
+    wellbore_marker_frame.from_marker_list(md_array = node_mds, marker_list = marker_list)
 
     # --------- Assert ----------
-    assert found_interpretation == horizon_interp_2
+    # Verify that attributes have been added correctly
+    assert wellbore_marker_frame.node_count == len(wellbore_marker_frame.node_mds) == len(
+        wellbore_marker_frame.marker_list) == 3
+    np.testing.assert_equal(node_mds, wellbore_marker_frame.node_mds)
```
andy-beer, line 233: As elsewhere, make sure to use floating point data and compare arrays with assert_array_almost_equal() from numpy.testing.

resqpy/well/_wellbore_marker_frame.py
kadijahassanali: something like this potentially? 
**title = str(self.trajectory.well_name) + '_' + marker.marker_type**

PR #341, opened 2021-12-09
Integration: add_surfaces bug - Remove deprecated call to crs_uuid

emmanesbit:
Remove deprecated call to crs_uuid in create_xml. This is now required at initialization stage (already in code)

### DIFF
--- a/resqpy/property/property_collection.py
+++ b/resqpy/property/property_collection.py
@@ -7,7 +7,7 @@ version = '1st December 2021'
 import logging
 
 log = logging.getLogger(__name__)
-log.debug('property.py version ' + version)
+log.debug('property_collection.py version ' + version)
 
 import numpy as np
 import numpy.ma as ma
--- a/resqpy/property/property_collection.py
+++ b/resqpy/property/property_collection.py
@@ -90,6 +90,7 @@ class PropertyCollection():
         ]  # list of (uuid, file_name, keyword, cached_name, discrete, uom, time_index, null_value,
         #                                   min_value, max_value, property_kind, facet_type, facet, realization,
         #                                   indexable_element, count, local_property_kind_uuid, const_value, points)
+        self.guessing = False
         if support is not None:
             self.model = support.model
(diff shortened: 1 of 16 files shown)


emmanesbit:
@andy-beer  @cflynn3  I've added the property.py debug fix to this PR too

resqpy/property/property_collection.py
```diff
@@ -90,6 +89,7 @@ def __init__(self, support = None, property_set_root = None, realization = None)
         ]  # list of (uuid, file_name, keyword, cached_name, discrete, uom, time_index, null_value,
         #                                   min_value, max_value, property_kind, facet_type, facet, realization,
         #                                   indexable_element, count, local_property_kind_uuid, const_value, points)
+        self.guessing = False
```
andy-beer, line 92: Should this be self.guess_warning = False ?

andy-beer (commented):
Looks good except for a suspected mistake in the new PropertyCollection attribute name.

resqpy/property/property_collection.py
emmanesbit: Good spot, just updated this now

resqpy/rq_import/_add_surfaces.py
```diff
@@ -116,7 +116,6 @@ def _add_single_surface(model, surf_file, surface_file_format, surface_role, qua
     surface.create_xml(ext_uuid,
                        add_as_part = True,
                        add_relationships = True,
-                       crs_uuid = rqet.uuid_for_part_root(crs_root),
```
andy-beer, line 119: Emma, Could you also please remove crs_root as an argument to this function _add_single_surface()

Then of course you will need to remove it in any calling code.  You can either make a breaking change by removing it from the argument list of the public equivalent function, or add a few lines there to convert to crs_uuid and produce a deprecation warning.

andy-beer (commented):
Just added another comment about crs_root

andy-beer (approved):
Good to merge into integration

PR #344, opened 2021-12-13
Moved check to set_tangent_vectors to the end of choose_init_method in Trajectory class init and re-na…

kadijahassanali:
- Shifted the check of whether tangent vectors should be calculated to the end of the __choose_init_method() method in Trajectory init. 
- Renamed "deviation_survey_file" argument to "ascii_trajectory_file"
- Added a **set_tangent_vectors = True** parameter to the **test_Trajectory_load_from_xml test** in resqpy/tests/test_trajectory.py

### DIFF
--- a/resqpy/well/_blocked_well.py
+++ b/resqpy/well/_blocked_well.py
@@ -955,7 +955,12 @@ class BlockedWell(BaseResqpy):
                 property_columns = df.columns[3:]
             self._add_df_properties(df, property_columns, length_uom = length_uom)
 
-    def import_from_rms_cellio(self, cellio_file, well_name, grid, include_overburden_unblocked_interval = False):
+    def import_from_rms_cellio(self,
+                               cellio_file,
+                               well_name,
+                               grid,
+                               include_overburden_unblocked_interval = False,
+                               set_tangent_vectors = True):
         """Populates empty blocked well from RMS cell I/O data; creates simulation trajectory and md datum.
 
         args:
@@ -963,6 +968,8 @@ class BlockedWell(BaseResqpy):
               exit points are expected
            well_name (string): the name of the well as used in the cell I/O file
            grid (grid.Grid object): the grid object which the cell indices in the cell I/O data relate to
+           set_tangent_vectors (boolean, default True): if True, tangent vectors will be computed from the well
+              trajectory's control points
 
         returns:
            self if successful; None otherwise
@@ -1050,7 +1057,7 @@ class BlockedWell(BaseResqpy):
                                                     length_uom,
                                                     well_name,
                                                     set_depth_zero = True,
-                                                    set_tangent_vectors = True)
+                                                    set_tangent_vectors = set_tangent_vectors)
 
                 self.node_count = len(trajectory_mds)
                 self.node_mds = np.array(trajectory_mds)
--- a/resqpy/well/_blocked_well.py
+++ b/resqpy/well/_blocked_well.py
@@ -1044,15 +1044,13 @@ class BlockedWell(BaseResqpy):
                                                                      well_name = well_name,
                                                                      grid_name = grid_name)
 
-                self.create_md_datum_and_trajectory(
-                    grid,
-                    trajectory_mds,
-                    trajectory_points,
-                    length_uom,
-                    well_name,
-                    set_depth_zero = True,
-                    set_tangent_vectors = False
-                )  # this is causing issues as the knot_count for the trajectory object is None and so cannot test that knot_count > 1
+                self.create_md_datum_and_trajectory(grid,
+                                                    trajectory_mds,
+                                                    trajectory_points,
+                                                    length_uom,
+                                                    well_name,
+                                                    set_depth_zero = True,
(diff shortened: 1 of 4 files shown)


kadijahassanali:
Hi @andy-beer, 

i made the recommended changes to the Trajectory classes' init method and included a "set_tangent_vectors = True" argument in the Trajectory_load_from_xml test. 
Thanks!

Best regards,
Kadija

resqpy/well/_blocked_well.py
```diff
@@ -1044,15 +1044,13 @@ def import_from_rms_cellio(self, cellio_file, well_name, grid, include_overburde
...
-                    set_depth_zero = True,
-                    set_tangent_vectors = False
-                )  # this is causing issues as the knot_count for the trajectory object is None and so cannot test that knot_count > 1
+                self.create_md_datum_and_trajectory(grid,
+                                                    trajectory_mds,
+                                                    trajectory_points,
+                                                    length_uom,
+                                                    well_name,
+                                                    set_depth_zero = True,
+                                                    set_tangent_vectors = True)
```
andy-beer, line 1053: Please add set_tangent_vectors as an argument to import_from_rms_cellio() and pass the value in from the BlockedWell __init__() method, rather than having it set to a constant value here.

resqpy/well/_trajectory.py
```diff
@@ -231,7 +229,10 @@ def __choose_init_method(self, set_tangent_vectors, data_frame, cell_kji0_list,
         elif wellspec_file:
             return 'wellspec_file'
         elif deviation_survey_file:
-            return 'deviation_survey_file'
+            return 'ascii_trajectory_file'
+        elif set_tangent_vectors and type(self.knot_count) is int and self.tangent_vectors is None:
+            if len(self.knot_count) > 1:
+                return 'tangent_vectors'
```
andy-beer, line 236: I would add a 'return None' line here to explicitly cover the situation when none of the conditions are met.  (Or raise an exception if it is an invalid situation, though here I think it should be allowed.)

resqpy/well/_trajectory.py
```diff
@@ -58,7 +58,7 @@ def __init__(
             cell_kji0_list = None,
             wellspec_file = None,
             spline_mode = 'cube',
-            deviation_survey_file = None,  # TODO: this file actually contains trajectory data --> rename argument!
+            ascii_trajectory_file = None,
             survey_file_space_separated = False,
```
andy-beer, line 62: The name of this argument should also be changed from survey_file... to ascii_file_space_separated or something like that. (also change the text in the docstring for this argument please).

resqpy/well/_trajectory.py
```diff
@@ -231,7 +229,10 @@ def __choose_init_method(self, set_tangent_vectors, data_frame, cell_kji0_list,
         elif wellspec_file:
             return 'wellspec_file'
         elif deviation_survey_file:
-            return 'deviation_survey_file'
+            return 'ascii_trajectory_file'
+        elif set_tangent_vectors and type(self.knot_count) is int and self.tangent_vectors is None:
```
andy-beer, line 233: Hmm, line 162 basically means that this set tangents logic will not be applied when loading from uuid (existing RESQML object).  As that was the whole point of having 'tangent_vectors' as an init option, we need to change something.  Today my feeling is strip that option out of the choose_init_method and init_function_dict after all.  We will treat the setting of tangents as an optional activity after whatever init method (including loading from uuid) has been applied.  In other words, let's put in a function call at lines 163 and 218 with that function applying the test that is currently here (line 233) and then calling self.set_tangents() if required.  Sorry to mess you about with a different steer every day!

tests/test_trajectory.py
```diff
@@ -209,7 +209,7 @@ def test_load_from_ascii_file(example_model_and_crs):
             # --------- Act ----------
             # Create a trajectory from the csv file
             trajectory_from_ascii = resqpy.well.Trajectory(parent_model = model,
-                                                           deviation_survey_file = trajectory_data_file_path,
+                                                           ascii_trajectory_file = trajectory_data_file_path,
```
andy-beer, line 212: We might want to set_tangent_vectors = True for one of these other tests and then check that the tangents are setup, as it will be taking a different route through the code.

andy-beer (changes requested):
@kadijahassanali, thanks for these changes which are mostly fine. Having had more of a look at the Trajectory initialisation code I am still not satisfied with the handling set_tangents (see comments).  Sorry!

resqpy/well/_trajectory.py
kadijahassanali: Not at all! This is why I've learnt more about resqpy in the last couple weeks than all the previous months I've been on the team.

resqpy/well/_trajectory.py
kadijahassanali: @andy-beer, I added the recommended 'return None' line to the __choose_init_method(). The try-except block  starting at line 201 should then log a warning if the combination of input arguments is invalid.

PR #346, opened 2021-12-15
Updated handling of const arrays and min/max values

emmanesbit:
Bug fix - updated handling of const arrays min/max values following refactoring


resqpy/property/_collection_create_xml.py
```diff
@@ -108,6 +107,7 @@ def _create_xml_patch_node(collection, p_node, points, const_value, indexable_el
 def _create_xml_property_min_max(collection, property_array, const_value, discrete, add_min_max, p_node, min_value,
                                  max_value):
     if add_min_max:
+        assert property_array is not None or const_value is not None
```
andy-beer, line 110: Should we allow a min_value and max_value to be passed through to here, without the need for property_array or const_value?  I think we can just scrap this assertion altogether.

andy-beer (approved):
I will test Patrick's code from this branch, so maybe hold off merging to master for a few minutes

PR #349, opened 2021-12-15
Organize CC reduction

NirjhorC:
Reducing cyclometric complexity in the Organize module

### DIFF
--- a/resqpy/organize/fault_interpretation.py
+++ b/resqpy/organize/fault_interpretation.py
@@ -109,28 +109,36 @@ class FaultInterpretation(BaseResqpy):
             return False
         if self is other or bu.matching_uuids(self.uuid, other.uuid):
             return True
-        if self.tectonic_boundary_feature is not None:
-            if not self.tectonic_boundary_feature.is_equivalent(other.tectonic_boundary_feature):
-                return False
-        elif other.tectonic_boundary_feature is not None:
+
+        attr_list = ['tectonic_boundary_feature', 'root', 'maximum_throw', 'mean_azimuth', 'mean_dip']
+        one_none_attr_list = [(getattr(self, v) is None) != (getattr(other, v) is None) for v in attr_list]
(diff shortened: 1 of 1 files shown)


resqpy/organize/fault_interpretation.py
```diff
@@ -109,28 +109,35 @@ def is_equivalent(self, other, check_extra_metadata = True):
...
+        if any(one_none_attr_list):
+            # If only one of self or other has a None attribute
             return False
-        if self.root is not None and other.root is not None:
-            if rqet.citation_title_for_node(self.root) != rqet.citation_title_for_node(other.root):
-                return False
+
+        # List of attributes that are not None in either self or other
+        non_none_attr_list = list(
+            filter(lambda v: not ((getattr(self, v) is None) and (getattr(other, v) is None)), attr_list))
```
andy-beer, line 121: As you have already checked that either both or neither are None, this line is unnecessarily repeating work with the **and** clause.  Please try to have some care for computational efficiency, even if it is not fashionable at the moment.

andy-beer: So, I don't think you need the filter and lambda at all. The following (untested) should do the trick unless I'm missing something:
`non_none_attr_list = [a for a in attr_list if getattr(self, a) is not None]`

andy-beer (commented):
Nirjhor, this is okay. Please take a look at my comments and either make a change as suggested or we can run with your code as it is if you feel strongly about it.

resqpy/organize/fault_interpretation.py
NirjhorC: You're right. The code would've returned False already if an attribute in only `self` or `other` was None. So at this point checking for None-ness (is that a word?) in `self` is sufficient.

andy-beer (approved):

cflynn3 (approved):

PR #350, opened 2021-12-15
Additional tests for gridpropertycollection

emmanesbit:


tests/test_property.py
```diff
@@ -1734,3 +1734,227 @@ def test_slice_for_box(example_model_with_properties):
...
+    numf = len(fine_pc.parts())
+
+    # Act
+    coarse_pc.extend_imported_list_copying_properties_from_other_grid_collection(other = fine_pc, coarsening = fc)
+    coarse_pc.write_hdf5_for_imported_list()
+    coarse_pc.create_xml_for_imported_list_and_add_parts_to_model()
+
+    # Assert
+    assert len(coarse_pc.parts()) == numc + numf
+    all_tens = np.zeros(shape = (3, 5, 5)) + 10
```
andy-beer, line 1759: You might like to use the numpy full() function instead of this construct:
all_tens = np.full(shape = (3, 5, 6), fill_value = 10)

andy-beer: PS. With the full() version, it would be slightly better to setup a float array, so fill_value = 10.0 or dtype = float

tests/test_property.py
```diff
@@ -1734,3 +1734,227 @@ def test_slice_for_box(example_model_with_properties):
...
+    # Arrange
+    model, coarse, fine, fc = example_fine_coarse_model
+
+    # Set up property collections
+    coarse_pc = rqp.GridPropertyCollection(grid = coarse)
+    numc = len(coarse_pc.parts())
+    fine_pc = rqp.GridPropertyCollection(grid = fine)
+
+    # Add a volume to the fine collection
+    inarray = np.zeros(shape = (6, 10, 10)) + 125  # fine grid dimensions are 5x5x5 so gross volume of 125
```
andy-beer, line 1776: np.full() comments apply here too

tests/test_property.py
```diff
@@ -1734,3 +1734,227 @@ def test_slice_for_box(example_model_with_properties):
...
+    assert np.min(ntg_array) > 0.4
+    assert np.max(ntg_array) < 0.7
+    assert np.allclose(np.mean(ntg_array), 0.550265)
+
+    # Check facies array
+    facies = [part for part in pc.parts() if pc.citation_title_for_part(part) == 'ab_facies'][0]
+    facies_array = pc.cached_part_array_ref(facies)
+    assert np.min(facies_array) == 0
+    assert np.max(facies_array) == 5
+    assert np.sum(facies_array) == 170
```
andy-beer, line 1960: If we ever regenerate the test file, this sum might need to be changed.  But quite good to keep it in as part of the test I feel.

tests/test_property.py
```diff
@@ -1734,3 +1734,227 @@ def test_slice_for_box(example_model_with_properties):
...
+    # Check NTG array
+    ntg = [part for part in pc.parts() if pc.citation_title_for_part(part) == 'ab_ntg'][0]
+    ntg_array = pc.cached_part_array_ref(ntg)
+    assert np.min(ntg_array) > 0.4
+    assert np.max(ntg_array) < 0.7
+    assert np.allclose(np.mean(ntg_array), 0.550265)
+
+    # Check facies array
+    facies = [part for part in pc.parts() if pc.citation_title_for_part(part) == 'ab_facies'][0]
+    facies_array = pc.cached_part_array_ref(facies)
```
andy-beer, line 1957: Could you please add a check that the facies part is flagged as discrete. I think  the following should do it:
assert not pc.continuous_for_part(facies)

andy-beer: ... and I guess something similar to check that the ntg part is continuous, above

andy-beer: Oh and whilst you are at it, why not check that the property kind is as expected too!

andy-beer (approved):
It's really good to see these tests being added!  I've suggested a few extra assertions for the ab import test, otherwise good to merge.

andy-beer (approved):

andy-beer (approved):

cflynn3 (approved):

PR #351, opened 2021-12-15
Refactoring olio\grid_functions and writing unit tests.

kadijahassanali:


kadijahassanali:
Hi @andy-beer,

I simplified two of the functions but focused on writing unit tests. The final complexity scores are shown below:
image

All functions except **columns_to_nearest_split_face()** have unit tests. The tests use very simple grids, so do let me know if I should use more complicated geometries.

Thanks!

resqpy/olio/grid_functions.py
```diff
@@ -175,10 +231,12 @@ def resequence_nexus_corp(corner_points, eight_mode = False, undo = False):
                         corner_points[k, j, i] = xyz.reshape((2, 2, 2, 3))
                     else:
                         xyz = corner_points[k, j, i].reshape((3, 8)).copy()
+                        print(xyz)
                         c = 0
                         for kp in range(2):
                             for jp in range(2):
                                 for ip in range(2):
+                                    print(corner_points[k, j, i, kp, jp, ip, :])
```
andy-beer, line 239: as above

resqpy/olio/grid_functions.py
```diff
@@ -481,6 +503,7 @@ def actual_pillar_shape(pillar_points, tolerance = 0.001):
     from_top = pp - pp[0]
     xy_drift = np.abs(from_top[:, :, 0]) + np.abs(
         from_top[:, :, 1])  # use Manhattan distance as cheap proxy for true distance
+    print(xy_drift)
```
andy-beer, line 506: is this also for testing / debugging?

resqpy/olio/grid_functions.py
```diff
@@ -545,45 +562,38 @@ def left_right_foursome(full_pillar_list, p_index):
...
-            return np.array([[False, True], [False, True]], dtype = bool)
-        elif exit == (0, 1):
-            return np.array([[False, False], [False, True]], dtype = bool)
-        else:
-            raise Exception('code failure whilst taking exit sides from dubious full pillar list')
-    else:
+    entry_tuples_list = [(0, 1), (0, -1), (1, 0), (-1, 0)]
+    exit_tuples_list = [[(-1, 0), (0, 1), (1, 0)], [(-1, 0), (0, -1), (1, 0)], [(0, -1), (1, 0), (0, 1)],
+                        [(0, -1), (-1, 0), (0, 1)]]
+    result_arrays_list = [[
```
andy-beer, line 568: Assuming this has been through yapf, you might like to disable its formatting of this list.  To do that add the following comment lines before and after the code to be spared reformatting:
```
# yapf: disable
# yapf: enable
```

tests/test_olio_grid_functions.py
```diff
@@ -0,0 +1,236 @@
...
+    depth[1, 2, 1] = 0.001  # below the default depth_zero_tolerance so should be set to 0 by function
+    depth[1, 2, 2] = 0.001
+    # Assertions to check that depth does have 2 values == 0.001
+    assert len(depth[depth == 0.001]) == 2
+    assert np.unique(thickness) == np.array([15.])
+    # --------- Act ----------
+    infill_block_geometry(extent = extent, depth = depth.transpose(), thickness = thickness, x = x, y = y)
+    # --------- Assert ----------
+    assert len(depth[depth == 0]) == 0
+    assert len(thickness[thickness == 22.5]) == 2
```
andy-beer, line 30: With snap_to_top_and_base defaulting to True, I don't think this is the answer we want here (though it might well be what the old code produced).  Let's discuss.

tests/test_olio_grid_functions.py
```diff
@@ -0,0 +1,236 @@
...
+
+def test_infill_block_geometry():
+
+    # --------- Arrange ----------
+    x_ = np.array([2, 4, 6, 8])
+    y_ = np.array([1, 3, 5, 7])
+    depth_ = np.array([10.0, 30.0, 50.0, 70.0])
+    x, y, depth = np.meshgrid(x_, y_, depth_)
+    extent = (4, 4, 4)
+    thickness = np.ones(extent) * 15
```
andy-beer, line 18: Personally I prefer to use `thickness = np.full(extent, 15)` when setting a numpy array to a constant value.

tests/test_olio_grid_functions.py
```diff
@@ -0,0 +1,236 @@
...
+                                   [[[0., 0., 10.], [10., 0., 10.]], [[0., -10., 10.], [10., -10., 10.]]]]]]])
+
+    # --------- Act ----------
+    resequence_nexus_corp(corner_points = corner_points)
+
+    # --------- Assert ----------
+    np.testing.assert_almost_equal(corner_points, expected_result)
+
+
+# TODO: Find out if grid is too simple
```
andy-beer, line 116: To make a significant test of the function, a complex geometry would be needed, with some cells pinched out and others collapsed in either the I or J direction, or any combination thereof.  At least this test will execute the function, so could pick up any basic errors not detected by flake8.

tests/test_olio_grid_functions.py
```diff
@@ -0,0 +1,236 @@
...
+    dxyz = (55.0, 65.0, 27.0)
+    grid = grr.RegularGrid(model, extent_kji = (4, 3, 2), title = 'concrete', origin = (0.0, 0.0, 1000.0), dxyz = dxyz)
+    grid.create_xml(add_cell_length_properties = True)
+    corner_points = grid.corner_points()
+
+    # --------- Act----------
+    (k, j, i) = random_cell(corner_points = corner_points)
+
+    # --------- Assert----------
+    assert k <= 4
```
andy-beer, line 133: The returned indices are zero based, so please make this assertion `0 <= k < 4`

Similarly for j and i below.

tests/test_olio_grid_functions.py
```diff
@@ -0,0 +1,236 @@
...
+    # --------- Assert----------
+    np.testing.assert_almost_equal(tri[0, :, :, 0][0][0], k_face_centre_points[0])
+    np.testing.assert_almost_equal(tri[0, :, :, 0][1][0], k_face_centre_points[1])
+    np.testing.assert_almost_equal(tri[1, :, :, 0][0][0], j_face_centre_points[0])
+    np.testing.assert_almost_equal(tri[1, :, :, 0][1][0], j_face_centre_points[1])
+    np.testing.assert_almost_equal(tri[2, :, :, 0][0][0], i_face_centre_points[0])
+    np.testing.assert_almost_equal(tri[2, :, :, 0][1][0], i_face_centre_points[1])
+
+
+def test_actual_pillar_shape(tmp_path):  # TODO: test with more complicated grid geometries
```
andy-beer, line 158: Yes, please add tests to cover the two other possible outcomes: 'straight' (but not vertical) and 'curved'.

I can give some suggestions about setting up other geometries – let's discuss.

tests/test_olio_grid_functions.py
```diff
@@ -0,0 +1,236 @@
...
+    # --------- Act----------
+    handedness_right = determine_corp_ijk_handedness(corner_points = corner_points_right)
+    handedness_left = determine_corp_ijk_handedness(corner_points = corner_points_left)
+
+    # --------- Assert----------
+    assert handedness_right == 'right'
+    assert handedness_left == 'left'
+
+
+# TODO: Find out if using a grid is too simple
```
andy-beer, line 193: I think this is good enough for testing this functionality.

tests/test_olio_grid_functions.py
```diff
@@ -0,0 +1,236 @@
...
+    [nk, nj, ni] = determine_corp_extent(corner_points = corner_points_reshaped)
+
+    # --------- Assert----------
+    assert [nk, nj, ni] == [4, 3, 2]
+
+
+@pytest.mark.parametrize(
+    'x_shift, y_shift, expected_result',
+    [(2, 0,
+      np.array([[[[[[[2.001, 0.001, 0.001], [12.001, 0.001, 0.001]], [[2.001, -10.001, 0.001], [12.001, -10.001, 0.001]]
```
andy-beer, line 216: You might like to use the yapf disable comment to allow the expected result arrays to be laid out in the same way as the corner_points array.

tests/test_olio_grid_functions.py
```diff
@@ -0,0 +1,236 @@
...
+    depth_ = np.array([10.0, 30.0, 50.0, 70.0])
+    x, y, depth = np.meshgrid(x_, y_, depth_)
+    extent = (4, 4, 4)
+    thickness = np.ones(extent) * 15
+
+    # Set two inactive cells
+    depth[1, 2, 1] = 0.001  # below the default depth_zero_tolerance so should be set to 0 by function
+    depth[1, 2, 2] = 0.001
+    # Assertions to check that depth does have 2 values == 0.001
+    assert len(depth[depth == 0.001]) == 2
```
andy-beer, line 24: I really dislike using == when comparing floating point numbers.  A value which has an exact decimal representation does not necessarily have the same exact value in the IEEE floating point representation.  A test that passes on one platform could fail elsewhere.  Therefore it is always better to use maths.isclose() or the numpy equivalent.

tests/test_olio_grid_functions.py
```diff
@@ -0,0 +1,236 @@
...
+    assert np.unique(thickness) == np.array([15.])
+    # --------- Act ----------
+    infill_block_geometry(extent = extent, depth = depth.transpose(), thickness = thickness, x = x, y = y)
+    # --------- Assert ----------
+    assert len(depth[depth == 0]) == 0
+    assert len(thickness[thickness == 22.5]) == 2
+
+
```
andy-beer, line 32: I think we should add some more testing of the infill_block_geometry() functionality at some point – doesn't need to be now!

andy-beer (commented):
Overall this looks good – let's discuss one or two of my comments.

resqpy/olio/grid_functions.py
kadijahassanali: Yup, I definitely should have removed all of those random print statements.

tests/test_olio_grid_functions.py
kadijahassanali: Hi @andy-beer,

I pinched out the entire layer and the expected cells are not returned by the random_cell function. I am clearly doing something wrong when I try to collapse one cell in the j direction, as that cellkji0 is still returned. This is an official SOS!

kadijahassanali:
Hi @andy-beer,

I didn't get around to testing columns_to_nearest_split_face. I saw examples of split faces in tests\test_fault_connection_set but was not 100% sure what they were doing. I'm happy to work on this when I get back on the 3rd if it's not too late then!
Thanks!

resqpy/olio/grid_functions.py
```diff
@@ -174,7 +230,7 @@ def resequence_nexus_corp(corner_points, eight_mode = False, undo = False):
                                     c += 1
                         corner_points[k, j, i] = xyz.reshape((2, 2, 2, 3))
                     else:
-                        xyz = corner_points[k, j, i].reshape((3, 8)).copy()
+                        xyz = corner_points[k, j, i].reshape((3, 8)).copy
```
andy-beer, line 233: surely the brackets are needed after .copy()

resqpy/olio/grid_functions.py
```diff
@@ -545,45 +559,39 @@ def left_right_foursome(full_pillar_list, p_index):
...
-    else:
+    # yapf: disable
+    entry_tuples_list = [(0, 1), (0, -1), (1, 0), (-1, 0)]
+    exit_tuples_list = [[(-1, 0), (0, 1), (1, 0)], [(-1, 0), (0, -1), (1, 0)], [(0, -1), (1, 0), (0, 1)],
+                        [(0, -1), (-1, 0), (0, 1)]]
+    result_arrays_list = [[
+        np.array([[False, True], [True, True]], dtype = bool),
+        np.array([[False, False], [True, True]], dtype = bool),
+        np.array([[False, False], [True, False]], dtype = bool)
+    ],
```
andy-beer, line 570: layout needs tidying up – that was the point of disabling yapf

tests/test_olio_grid_functions.py
```diff
@@ -0,0 +1,306 @@
...
+                           dxyz = dxyz,
+                           as_irregular_grid = True)
+
+    grid_points = grid.points_ref(masked = False)
+
+    # pinch out cells in the k == 1 layer
+    for j in range(3):
+        for i in range(3):
+            grid_points[2][j][i][-1] = 1100.
+    grid.write_hdf5()
```
andy-beer, line 140: The write_hdf5(), create_xml(), store_epc() should happen later, after all modifications of grid_points

tests/test_olio_grid_functions.py
```diff
@@ -0,0 +1,306 @@
...
+    grid.create_xml(add_cell_length_properties = True)
+    grid_uuid = grid.uuid
+    model.store_epc()
+
+    # collapse cellkji0 == 0,0,0 in the j direction # TODO: figure out why this does not work!
```
andy-beer, line 145: The changes to grid_points made here are not preserved, due to the write_hdf5() call already having been made.

tests/test_olio_grid_functions.py
```diff
@@ -0,0 +1,306 @@
...
+
+    # --------- Assert----------
+    assert [nk, nj, ni] == [4, 3, 2]
+
+
+@pytest.mark.parametrize(
+    # yapf: disable
+    'x_shift, y_shift, expected_result',
+    [(2, 0,
+      np.array([[[[[[[2.001, 0.001, 0.001], [12.001, 0.001, 0.001]], [[2.001, -10.001, 0.001], [12.001, -10.001, 0.001]]
```
andy-beer, line 285: layout of array to be tidied up – that is the point of yapf disable

andy-beer (changes requested):
Some small changes needed – I'll probably do those myself

tests/test_olio_grid_functions.py
andy-beer: As indicated in code comments, I believe the issue is a misplaced write_hdf5().  I will make the correction.

tests/test_olio_grid_functions.py
```diff
@@ -0,0 +1,306 @@
...
+        assert 0 <= j < 2
+        assert 0 <= i < 2
+        # check that none of the k,j,i combinations that correspond to pinched cells are chosen by the random_cell function
+        assert (k, j, i) not in [(1, 0, 0), (1, 0, 1), (1, 1, 0), (1, 1, 1)]
+        # TODO: figure out why is cell kji0 == (0, 0, 0) still returned?
+        # print(k, j, i)
+        trial_number += 1
+    np.testing.assert_almost_equal(new_extent,
+                                   np.array([3, 2, 2
+                                            ]))  #TODO: confirm this makes sense if an entire k==1 layer is pinched out
```
andy-beer, line 175: Yes, a pinched out layer still exists as a layer in the logical KJI space, so the grid extent is unchanged.

PR #355, opened 2021-12-16
Unittests and Bugfix for Organize

NirjhorC:
Early PR to fix a bug introduced during the refactoring flagged by some new unittests

### DIFF
--- a/resqpy/organize/fault_interpretation.py
+++ b/resqpy/organize/fault_interpretation.py
@@ -121,14 +121,14 @@ class FaultInterpretation(BaseResqpy):
 
         # Additional tests for attributes that are not None
         check_dict = {
-            'tectonic_boundary_feature': self.tectonic_boundary_feature.is_equivalent(other.tectonic_boundary_feature),
-            'root': rqet.citation_title_for_node(self.root) != rqet.citation_title_for_node(other.root),
-            'maximum_throw': maths.isclose(self.maximum_throw, other.maximum_throw, rel_tol = 1e-3),
-            'mean_azimuth': maths.isclose(self.mean_azimuth, other.mean_azimuth, abs_tol = 0.5),
-            'mean_dip': maths.isclose(self.mean_dip, other.mean_dip, abs_tol = 0.5)
+            'tectonic_boundary_feature': lambda: self.tectonic_boundary_feature.is_equivalent(other.tectonic_boundary_feature),
+            'root': lambda: rqet.citation_title_for_node(self.root) != rqet.citation_title_for_node(other.root),
+            'maximum_throw': lambda: maths.isclose(self.maximum_throw, other.maximum_throw, rel_tol = 1e-3),
+            'mean_azimuth': lambda: maths.isclose(self.mean_azimuth, other.mean_azimuth, abs_tol = 0.5),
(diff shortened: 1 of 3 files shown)


cflynn3 (approved):

resqpy/organize/fault_interpretation.py
```diff
@@ -121,14 +121,19 @@ def is_equivalent(self, other, check_extra_metadata = True):
...
+            'maximum_throw':
+                lambda: maths.isclose(self.maximum_throw, other.maximum_throw, rel_tol = 1e-3),
+            'mean_azimuth':
+                lambda: maths.isclose(self.mean_azimuth, other.mean_azimuth, abs_tol = 0.5),
+            'mean_dip':
+                lambda: maths.isclose(self.mean_dip, other.mean_dip, abs_tol = 0.5)
         }
 
-        check_outcomes = [check_dict[v] for v in non_none_attr_list]
+        check_outcomes = [check_dict[v]() for v in non_none_attr_list]
```
andy-beer, line 136: I haven't seen this combination of dictionary and lambda functions – I trust it works!

tests/test_fault_interpretation.py
```diff
@@ -0,0 +1,142 @@
...
+    result = fault_interp.is_equivalent(other = other)
+
+    # Assert
+    assert result is True
+
+
+def test_is_equivalent_is_sameUUIDs(tmp_model):
+
+    # Arrange
+    object_uuid = uuid.uuid4()
```
andy-beer, line 57: It might be better to use `object_uuid = resqpy.olio.new_uuid()` instead.

tests/test_fault_interpretation.py
```diff
@@ -0,0 +1,142 @@
...
+        title = 'test_fault_interpretation',
+    )
+    fault_interp.tectonic_boundary_feature = tect_boundary
+
+    # Act
+    tect_boundary_other = rqo.TectonicBoundaryFeature(tmp_model, kind = 'fracture')
+
+    other = rqo.FaultInterpretation(
+        tmp_model,
+        title = 'test_fault_interpretation_other',
```
andy-beer, line 113: Please change this title so that both titles are the same – that way we are sure it is the tectonic boundary feature that is causing the non-equivalent result.

andy-beer (changes requested):
This is good but could you please make the changes to some of the titles in the test_fault_interpretation.py module (see comments). Thanks.

tests/test_fault_interpretation.py
NirjhorC: Done

tests/test_fault_interpretation.py
NirjhorC: Done

andy-beer: Nirjhor, Have you committed your latest changes? The titles are still showing as different when I view code changes in github.

andy-beer (changes requested):
Nirjhor, I think there are some other instances in your _is_equivlanet()_ tests where you are setting the title for one object to X and for the other to X_other.  That is fine if the test wants to check the behaviour of different titles but if testing that differences in other attributes are detected, the titles should be set the same.

andy-beer (approved):
Thanks Chris, good to merge.

PR #356, opened 2021-12-17
Tests - Single additional test to cover properties on gridd2drepresentation objects

emmanesbit:
Single test to cover property collections with a grid2drepresentation as a support. Properties with indexable of cells and nodes added for testing.

### DIFF
--- a/tests/test_property.py
+++ b/tests/test_property.py
@@ -2036,3 +2036,63 @@ def test_remove_cached_from_imported_list(example_model_with_properties):
     pc.remove_cached_imported_arrays()
     # Assert
     assert not hasattr(pc, array_name)
+
+
+def test_mesh_support(example_model_and_crs):
+    # Arrange
+    model, crs = example_model_and_crs
+    # create some random depths
+    z = (np.random.random(3 * 3) * 20.0 + 1000.0).reshape((3, 3))
+    # Create some properties
+    cell_prop = np.full(shape = (2, 2), fill_value = 2)
+    cell_prop[:, 0] = 1
+    node_prop = np.full(shape = (3, 3), fill_value = 10)
+    node_prop[:, 0] = 0
+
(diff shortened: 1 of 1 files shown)


cflynn3 (approved):

tests/test_property.py
```diff
@@ -2036,3 +2036,63 @@ def test_remove_cached_from_imported_list(example_model_with_properties):
...
+    pc = rqp.PropertyCollection()
+    pc.set_support(support = mesh)
+
+    # Assert
+    assert pc.support is not None
+    assert isinstance(pc.support, rqs.Mesh)
+    assert pc.support == mesh
+    assert pc.support_uuid == mesh.uuid
+
+    # Act - add different indexable element properties to collection
```
cflynn3, line 2077: Ideally these would be separate tests, as they represent separate test cases.

tests/test_property.py
```diff
@@ -2036,3 +2036,63 @@ def test_remove_cached_from_imported_list(example_model_with_properties):
...
+                                         indexable_element = 'nodes')
+    pc.write_hdf5_for_imported_list()
+    pc.create_xml_for_imported_list_and_add_parts_to_model()
+
+    # Assert
+    for part in pc.parts():
+        if pc.citation_title_for_part(part) == 'TESTnode':
+            assert pc.indexable_for_part(part) == 'nodes'
+        else:
+            assert pc.indexable_for_part(part) == 'cells'
```
andy-beer, line 2098: Could be good to put in a few more assertions. For example to call the pc.cached_part_array_ref() method and check the shape (and maybe the values) of the array.

andy-beer (commented):
This is good Emma – could you add a few more assertions at the end there please.

andy-beer (approved):
Ready to merge

PR #360, opened 2021-12-17
testing of Model catalogue functions and a few other methods

andy-beer:


andy-beer:
I think this batch of tests is worth merging in, even though the Model._forestry module is not covered yet.

cflynn3 (approved):

PR #369, opened 2022-01-06
Olio/write_faults.py tests added

maxm17:
Olio/write_faults.py tests added

### DIFF
--- a/tests/test_olio_write_faults.py
+++ b/tests/test_olio_write_faults.py
@@ -1,0 +1,90 @@
+import pandas as pd
+import pytest
+import resqpy.olio.write_faults as wf
+from pytest_mock import MockerFixture
+
+
+@pytest.mark.parametrize("test_df", [pd.DataFrame(), None])
+def test_no_data_in_faults(test_df):
+    # Act & Assert
+    with pytest.raises(AssertionError):
+        wf.write_faults_nexus('test', test_df)
+
+
+def test_writing_faults_to_file(mocker: MockerFixture):
+    # Arrange
+    test_df = pd.DataFrame({
+        'name': ['fault1', 'fault2'],
+        'i1': [720, 305],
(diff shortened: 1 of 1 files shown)


cflynn3 (approved):

tests/test_olio_write_faults.py
```diff
@@ -0,0 +1,90 @@
+import pandas as pd
+import pytest
+import resqpy.olio.write_faults as wf
+from pytest_mock import MockerFixture
+
+
+@pytest.mark.parametrize("test_df", [pd.DataFrame(), None])
+def test_no_data_in_faults(test_df):
+    # Act & Assert
+    with pytest.raises(AssertionError):
```
connortann, line 10: I think a `ValueError` would be more appropriate than an `AssertionError` here: asserts are generally used to aid debugging internal logic, and shouldn't be raised by users.

connortann (changes requested):
Suggest small change but otherwise LGTM

tests/test_olio_write_faults.py
connortann: Following a chat with @cflynn3  - let's park this change for a separate PR. If we change from asserts to Exceptions, we might as well do that consistently across the code.

connortann (approved):

tests/test_olio_write_faults.py
maxm17: I'm testing that the code produces an AssertionError. Do you want me to alter the write_faults.py file to produce a value error?

connortann: That was my initial suggestion yes - but I think it's ok as it is for now, we could make that change at a later date.

maxm17: I agree - I also want to add type hints and more detailed docstrings too

PR #372, opened 2022-01-06
Olio/utility.py tests added

maxm17:


connortann:
It looks like the `yamf` auto-formatter hasn't run, so the linting job is failing.

If you use VS Code, code formatting can be applied as a setting to always run when you press "save", which is a neat time saver.

maxm17:
Yeah I'll get that setup now!

cflynn3 (approved):

PR #376, opened 2022-01-10
Olio/write_data.py tests added. Slight refactor to other files.

maxm17:


tests/test_olio_write_data.py
```diff
@@ -0,0 +1,286 @@
...
+
+def test_write_pure_binary_data(mocker: MockerFixture, caplog):
+    # Arrange
+    test_array = np.array([[[0, 0], [0, 0]], [[0, 0], [0, 0]]])
+    open_mock = mocker.mock_open()
+    fileno_mock = mocker.Mock(return_value = 1, name = 'fileno_mock')
+    open_mock.return_value.fileno = fileno_mock
+
+    mocker.patch("builtins.open", open_mock)
+    expected_calls = [mocker.call('test', 'wb'), mocker.call().__enter__()]
```
cflynn3, line 15: You should check that these calls are actually called in the assertions.

cflynn3 (changes requested):
Please add the assertion to test_write_pure_binary_data, otherwise this looks good to me.

PR #377, opened 2022-01-11
veronoi function refactored

maxm17:
veronoi function refactored. Naming of inner functions need to be checked

### DIFF
--- a/resqpy/olio/triangulation.py
+++ b/resqpy/olio/triangulation.py
@@ -16,13 +16,14 @@ import resqpy.model as rq
 import resqpy.olio.intersection as meet
 import resqpy.olio.vector_utilities as vec
 
+
 # _ccw_t() no longer needed: triangle vertices maintained in anti-clockwise order throughout
 # def _ccw_t(p, t):   # puts triangle vertex indices into anti-clockwise order, in situ
 #    if vec.clockwise(p[t[0]], p[t[1]], p[t[2]]) > 0.0:
 #       t[1], t[2] = t[2], t[1]
 
 
-def _dt_simple(po, plot_fn = None, progress_fn = None, container_size_factor = None):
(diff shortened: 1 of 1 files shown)


resqpy/olio/triangulation.py
```diff
@@ -658,7 +676,7 @@ def triangulated_polygons(p, v, centres = None):
     t_index = 0
 
     for cell, poly_vertices in enumerate(v):
-        # add polygon centre points to points array, if not already provided
+        # add polygon centre points to point array, if not already provided
```
cflynn3, line 679: I think 'points' was correct, as in an array of points?

emmanesbit (approved):
Looks good to me

resqpy/olio/triangulation.py
andy-beer: yes, 'points' is better

andy-beer:
I see that the branch has been merged into master.  Did we check that the code changes did not change the functionality?  ie. Are we confident that the same Voronoi diagram is being computed for the same input, at least for a few sets of input data?

maxm17:
@andy-beer Hi Andy, I didn't test the output with any real data, I just used the unit tests that were already in place which had good coverage

andy-beer:
Thanks for the clarification Max @maxm17 .  We have to be a bit careful when relying on the coverage statistics.  Those are a measure of what proportion of the lines of code have been executed by the unit tests.  The unit tests might _**or might not**_ check that the functionality is correct, ie. whether the code gets the right answer.

In the case of Delauney triangulation and the computation of Voronoi diagrams, the unit tests can only make a few very basic sanity assertions about the answers.  They do not check that the answers are correct.  This is because to do so would require the unit tests to have another implementation of the complex calculation of these functions, which is not practical.

Instead, when writing the original code, I augmented the automated testing with manual testing which involved me literally looking at the diagrams being generated. Visually a human eye can determine whether the answer is correct or not without too much trouble.

PR #384, opened 2022-01-18
Change python packaging system from setuptools to Poetry

### LINKED ISSUE #379: Fix dependency managment and setup instructions for the project
- [x] Move the requirements.txt file to the base folder
- [x] Update the dependencies to the correct versions
- [x] Update the documentation to use PIP and venv rather than conda
- [ ] Update the build server to run using the correct dependencies
- [x] Add instructions for keeping dev environments up to date
- [x] Separate Dev and Package dependencies 
- [x] Test the new instructions on Linux, Windows and Mac

cflynn3:
Closes #379


connortann:
This looks promising - I'm not sure it's ready for merging, as I think there are still hangovers of the previous packaging setup. In particular, the CI jobs still use the setuptools-based approach for building and deploying.

As I understand, python poetry removes the need to maintain a setup.cfg / setup.py. To avoid being in a messy "half-way house", I think if we're using poetry for packaging we should switch the build and deploy steps to use Poetry too, and remove the setup.cfg information. This would mean moving the package metadata from setup.cgf to pyproject.toml , and also editing the github CI workflows.

connortann:
Also, before proceeding I think we need a clear plan for versioning. Python Poetry manages the version using pyproject.toml, which is inconsistent with the current setup using setuptools-scm . I do not know if it is even possible to get poetry working with versions from git, which may be a blocker.

Some possible workarounds here: https://www.nicholasnadeau.com/post/2020/8/one-version-to-rule-them-all-keeping-your-python-package-version-number-in-sync-with-git-and-poetry/

connortann:
Further to the above, this could be one Poetry-compatible way of ensuring we still have PEP-440 compliant version strings that are single-sourced from version control:

https://pypi.org/project/poetry-dynamic-versioning/

connortann (changes requested):
As per comments, I think there are still setuptools remnants, and the migration to poetry is not complete.

pyproject.toml
```diff
@@ -5,10 +5,39 @@
...
 write_to = "resqpy/version.py"
+
+[tool.poetry]
+name = "resqpy"
+version = "1.6.1"
+description = "Python API for working with RESQML models"
+authors = ["BP"]
+license = "MIT"
+
+[tool.poetry.dependencies]
```
connortann, line 23: These dependencies seem far too restrictive. Specification of exact versions is handled in the Poetry.lock file, and I do not think we should reproduce that here.

I believe this section should contain version _constraints_ only, not restricting all users to only use these exact versions. I would suggest using the Poetry style of "caret requirements":
https://python-poetry.org/docs/dependency-specification/

.github/workflows/ci-publish.yml
```diff
@@ -57,7 +57,10 @@ jobs:
       - name: Install from build
         run: |
           python -m pip install --upgrade pip
-          pip install resqpy[tests] --find-links=dist/
+          pip install poetry
+          pip install poetry-dynamic-versioning
+          poetry build
+          # pip install resqpy[tests] --find-links=dist/
```
connortann, line 63: I think this step should be unchanged. Poetry should build the package as a python wheel in a previous step - this  should be pip-installable in the usual way here.

.github/workflows/ci-publish.yml
```diff
@@ -84,6 +87,6 @@ jobs:
           TWINE_USERNAME: ${{ secrets.PYPI_USERNAME }}
           TWINE_PASSWORD: ${{ secrets.PYPI_PASSWORD }}
         run: |
-          python -m pip install --upgrade pip
+          poetry build
           pip install twine
           twine upload dist/*
```
connortann, line 92: I believe poetry replaces the need for twine, so I think we should be using "poetry publish"

.github/workflows/ci-tests.yml
```diff
@@ -26,7 +26,13 @@ jobs:
...
         run: |
           python -m pip install --upgrade pip
-          pip install -e ".[tests]"
+          pip install poetry
+          pip install poetry-dynamic-versioning
+          pip install flake8
+          pip install mypy
+          pip install yapf
+          pip install toml
+          poetry install
```
connortann, line 35: Poetry should install the dev dependencies as specified in our pyproject.toml, so we should not need to duplicate them here

.github/workflows/ci-tests.yml
```diff
@@ -70,12 +76,13 @@ jobs:
 
       - name: Install resqpy
         run: |
-          python -m pip install --upgrade pip setuptools wheel
-          pip install -e ".[tests]"
+          pip install poetry
+          pip install poetry-dynamic-versioning
+          poetry install
```
connortann, line 81: These steps look redundant from step above. Suggest this step just installs resqpy, and previous step just installs poetry.

pyproject.toml
connortann: I think same comment for dev dependencies below?

setup.cfg
```diff
@@ -19,38 +19,10 @@ classifiers =
     Topic :: System :: Filesystems
```
connortann, line 19: I believe this packaging information is for setuptools, so should be migrated to poetry's pyproject.toml

.github/workflows/ci-tests.yml
cflynn3: The unit tests run in parallel on a separate build agent to the static analysis task, therefore poetry needs to be installed on both.

.github/workflows/ci-tests.yml
cflynn3: Unfortunately the linter Github action doesn't run from within the poetry environment, so we need to install these dependencies separately.

.github/workflows/ci-publish.yml
```diff
@@ -84,6 +87,5 @@ jobs:
           TWINE_USERNAME: ${{ secrets.PYPI_USERNAME }}
           TWINE_PASSWORD: ${{ secrets.PYPI_PASSWORD }}
         run: |
-          python -m pip install --upgrade pip
-          pip install twine
-          twine upload dist/*
+          poetry build
```
connortann, line 90: Suggest removing this build command, as the package should already been built in a previous step. It may be worth checking that poetry uses the same `dist` directory for the python wheels.

.github/workflows/ci-publish.yml
```diff
@@ -84,6 +87,5 @@ jobs:
           TWINE_USERNAME: ${{ secrets.PYPI_USERNAME }}
           TWINE_PASSWORD: ${{ secrets.PYPI_PASSWORD }}
```
connortann, line 88: As we are using poetry rather than twine, this injection of PyPI credentials needs to be updated

.github/workflows/ci-tests.yml
connortann: Can this be rectified with the `poetry shell` command?

.github/workflows/ci-tests.yml
connortann: Yes indeed you're right, I don't know what I was thinking yesterday!

.github/workflows/ci-tests.yml
cflynn3: No, shell doesn't work on Github's CI/CD servers.

setup.cfg
```diff
@@ -1,56 +1,7 @@
...
-install_requires = 
-    numpy
-    pandas
-    h5py
-    lxml
-    lasio
-
 [options.package_data]
 # Include resqpy/olio/data/*.json
 * = data/*.json
```
connortann, line 3: I believe this section is also setuptools-specific, so can be removed / migrated. The equivalent in poetry I think is the "Include" section:

https://python-poetry.org/docs/pyproject/#include-and-exclude

tests/unit_tests/test_resqpy.py
```diff
@@ -24,10 +24,8 @@ def test_all_imports():
...
 def test_version():
     # This is dynamically created when package is built
-    # If this fails when running tests locally, ensure you have installed the dev dependencies specified in setup.cfg
-    # In particular, try:  pip install setuptools_scm
     version_string = resqpy.__version__
 
     # Ensure version string is a PEP-440 compliant version
     version = Version(version_string)
-    assert version >= Version("1.1.1")
+    assert version >= Version("0.0.0")
```
connortann, line 31: I'm a bit hesitant to disable this test: given that we've had some versioning issues in the past, it would give us some peace of mind to know the new system works.

I believe the version from Poetry-dynamic-versioning will only be present when the package is built (see [this thread](https://github.com/mtkennerly/poetry-dynamic-versioning/issues/25)). The CI tests are run against the built package, so it should be fine in that context.

However, we probably want to skip this test when running tests locally, as most devs will not have built the package.

One option for implementing this would be to create a custom `@requirebuild` decorator as detailed in [this thread](https://stackoverflow.com/questions/33084190/default-skip-test-unless-command-line-parameter-present-in-py-test), which disables the test unless the `--requirebuild` flag is passed to pytest.

.github/workflows/ci-publish.yml
```diff
@@ -54,13 +55,15 @@ jobs:
...
+          pip install poetry
+          pip install poetry-dynamic-versioning
+          poetry install
 
       - name: Checkout code
         uses: actions/checkout@v2
 
       - name: Run tests
-        run: pytest
+        run: poetry run pytest
```
connortann, line 66: Continuing the other thread about tests, here is perhaps where we could use a `--requirebuild` flag to enable extra tests, specifically the one that checks the version number is nonzero.

.github/workflows/ci-tests.yml
```diff
@@ -26,7 +26,13 @@ jobs:
       - name: Install dependencies
         run: |
           python -m pip install --upgrade pip
-          pip install -e ".[tests]"
+          pip install poetry
```
connortann, line 29: Suggest we change the way that poetry is installed into the CI environment.

Many threads suggests that we'll have issues if poetry is pip-installed in the same environment as a package, for example: https://github.com/python-poetry/poetry/issues/2102

The recommended way to install poetry in a GitHub action seems to be with this action:
https://github.com/snok/install-poetry

This has added bonus of enabling caching of dependencies, which should speed up our CI pipeline significantly.

.github/workflows/ci-tests.yml
connortann: It looks like `source .venv/bin/activate` is an alternative (from the README of https://github.com/snok/install-poetry )

.github/workflows/ci-tests.yml
```diff
@@ -70,12 +89,13 @@ jobs:
 
       - name: Install resqpy
         run: |
-          python -m pip install --upgrade pip setuptools wheel
-          pip install -e ".[tests]"
+          pip install poetry
```
connortann, line 92: It looks like this step uses a different method to install & configure Poetry than the "static analysis" one. For consistency, could we make them the same?

Given we will need to repeat several steps (install poetry, dependencies, caching etc), I think we could reduce duplication by using YAML anchors & aliases.

.github/workflows/ci-tests.yml
```diff
@@ -23,31 +25,48 @@ jobs:
...
 
-      # Post in-line comments for any issues found
-      # Do not run if coming from a forked repo
-      # See https://github.com/marketplace/actions/lint-action
       - name: Run linters (with annotations)
         if: github.event.pull_request.head.repo.full_name == github.repository
         uses: wearerequired/lint-action@v1
         with:
           flake8: true
+          flake8_command_prefix: poetry run
```
connortann, line 55: Nice! Very clean way to get this linting job working with our env

.github/workflows/ci-publish.yml
```diff
@@ -51,16 +57,23 @@ jobs:
...
+        with:
+          virtualenvs-create: true
+          virtualenvs-in-project: true
+
       - name: Install from build
         run: |
           python -m pip install --upgrade pip
-          pip install resqpy[tests] --find-links=dist/
+          pip install poetry-dynamic-versioning
+          poetry install
```
connortann, line 70: Should we use the existing `pip install resqpy --find-links=dist/` here, to ensure we are running the tests against the built package?

This step was originally intended to ensure that the python wheel functions correctly and can be pip-installed by end users.

.github/workflows/ci-tests.yml
```diff
@@ -23,31 +25,48 @@ jobs:
...
+
+      - name: Install missing cached dependencies
+        if: steps.cached-poetry-dependencies.outputs.cache-hit != 'true'
+        run: poetry install --no-interaction
+
+      - name: Install other dependencies
         run: |
           python -m pip install --upgrade pip
-          pip install -e ".[tests]"
+          pip install poetry-dynamic-versioning
```
connortann, line 48: I think we should use `poetry plugin add` rather than `pip install` here, as that ensures the plugin is installed into the poetry env and not the project virtual env.

Similarly, I think this command would logically make sense as a separate step immediately after "install poetry". Same for other CI jobs that install poetry. 

More info on that command here:
https://python-poetry.org/docs/master/plugins/

connortann:
FYI, I've made a PR on the "Install Poetry" GitHub Action to allow it to install plugins as well. However, would recommend waiting until Poetry 1.2 is officially released before using this.

https://github.com/snok/install-poetry/pull/69

.github/workflows/ci-tests.yml
cflynn3: This might be possible on future versions of Poetry, but it's not currently supported in the latest 'stable' version.

.github/workflows/ci-tests.yml
connortann: Ah, I see that YAML anchors & aliases aren't actually supported in GitHub Actions. Shame!

An alternative is [Reusable workflows](https://docs.github.com/en/actions/using-workflows/reusing-workflows), which seems to be essentially putting the common steps (Setup python, install poetry, install plugins, load deps from cache) into a separate yml workflow and then calling that as a step. Just a suggestion.

connortann (approved):
LGTM! Thanks for addressing all of my comments. @andy-beer , FYI this changes the release pipeline significantly so would be good to ensure the release process still works after this is merged.

PR #390, opened 2022-01-20
appending realisation number to Nexus grid property file names

andy-beer:
When exporting grid properties to Nexus ascii files, the filenames need to include a realisation number if the property collection includes multiple realisations.


andy-beer:
@cflynn3 This pull request is mainly to add a realisation number to ascii file names generated when exporting grid properties ready for Nexus. However, it also includes a fix for a bug in the Grid module write_nexus_corp() method.  This method is in use in existing workflows, so would be good to get resolved.

cflynn3 (approved):

PR #392, opened 2022-01-21
cell_properties tests added

maxm17:


tests/unit_tests/grid/test_cell_properties.py
```diff
@@ -0,0 +1,281 @@
...
+    if hasattr(grid, 'array_thickness'):
+        delattr(grid, 'array_thickness')
+    if hasattr(grid, 'property_collection'):
+        delattr(grid, 'property_collection')
+
+    # Act
+    thickness = cp.thickness(grid)
+
+    # Assert
+    np.testing.assert_array_almost_equal(thickness, 20.0)
```
cflynn3, line 125: Surprised assert_array_almost_equal works comparing against a float?

tests/unit_tests/grid/test_cell_properties.py
```diff
@@ -0,0 +1,281 @@
...
+    if hasattr(grid, 'array_volume'):
+        delattr(grid, 'array_thickness')
+    if hasattr(grid, 'property_volume'):
+        delattr(grid, 'property_collection')
+
+    # Act
+    volume = cp.volume(grid)
+
+    # Assert
+    np.testing.assert_array_almost_equal(volume, 100000.0)
```
cflynn3, line 231: As above, I think a different assert method should be used.

cflynn3 (changes requested):
Looks good, but please fix the two highlighted assert methods.

tests/unit_tests/grid/test_cell_properties.py
andy-beer: numpy has the concept of 'array like' which will take a scalar and treat it as a constant array when needed, if I've understood correctly, so I prefer the form Max has used here

cflynn3: OK, in that case I retract my comments :-).

cflynn3 (approved):

PR #393, opened 2022-01-21
WIP: correcting well package init for docstring processing to readthedocs

### LINKED ISSUE #389: Readthedocs for well package missing

andy-beer:
closes #389


connortann:
Looks good, just need to rerun yapf.

To prevent this kind of hassle in future, we could investigate using "pre-commit hooks" so that flake8 and yapf are automatically run locally before commits?

PR #397, opened 2022-01-24
adding test data for small ensemble of Nexus output vdb's

andy-beer:
This change adds some test data in Nexus vdb output format (along with a few of the Nexus input files and a RESQML dataset used to generate those input files). There are 3 realisations, each with different grid properties.

The main purpose of this test data is to allow unit testing of resqpy import from vdb.

Nexus is a trademark of Halliburton.


tests/test_data/wren/ACTIVE
```diff
@@ -0,0 +1,33 @@
+! Data written by write_array_to_ascii_file() python function
+! Extent of array is: [8, 5, 3]
+! Maximum 20 data items per line
+
+1	1	1	1	1	1	1	1
+
```
connortann, line 6: Just to confirm, should this file be "double spaced" with blank lines between the rows?

andy-beer: I noticed that too. It is okay. (It is a Nexus input file and Nexus ignores those blank lines.) It was generated from resqpy with an option that puts blank lines between blocks of data (based on change in J index). Most real models would have a big batch of lines followed by a blank, which can make it easier to find a block of data.  In this case, NI is tiny (8), so all the data for one block sits on one line.

connortann (approved):
LGTM

Issue #401, opened 2022-01-27
set_face_set_gcs_list_from_dict function bug


maxm17:
When unpacking face_set_dict:

kelp_j, kelp_i, axis = face_set_dict[feature]
ValueError: not enough values to unpack (expected 3, got 2)

PR #408, opened 2022-01-31
defensive coding fix for set_face_set_gcs_list_from_dict function

andy-beer:
resolves issue #401

### DIFF
--- a/resqpy/grid/face_functions.py
+++ b/resqpy/grid/face_functions.py
@@ -99,7 +99,7 @@ def set_face_set_gcs_list_from_dict(grid, face_set_dict = None, create_organizin
             kelp_j, kelp_i = kelp_values
             axis = 'K'
         elif len(kelp_values) == 3:
-            kelp_j, kelp_i, axis = face_set_dict[feature]
+            kelp_j, kelp_i, axis = kelp_values
         else:
             raise Exception('grid face set dictionary item messed up')
         kelp_j, kelp_i, axis = face_set_dict[feature]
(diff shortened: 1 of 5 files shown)


resqpy/grid/face_functions.py
```diff
@@ -93,8 +93,15 @@ def set_face_set_gcs_list_from_dict(grid, face_set_dict = None, create_organizin
...
-    for feature in face_set_dict:
+    for feature, kelp_values in face_set_dict.items():
         gcs = rqf.GridConnectionSet(grid.model, grid = grid)
+        if len(kelp_values) == 2:
+            kelp_j, kelp_i = kelp_values
+            axis = 'K'
+        elif len(kelp_values) == 3:
+            kelp_j, kelp_i, axis = face_set_dict[feature]
+        else:
+            raise Exception('grid face set dictionary item messed up')
```
maxm17, line 104: I would change the Exception to a ValueError

maxm17 (changes requested):

andy-beer:
Thanks for those comments @maxm17 – all 3 changes made as suggested.

andy-beer:
Max, I couldn't find the unit test that had flagged up this bug.  Maybe I was looking in the wrong branch.  Anyway, could you please track down the test and run it against this branch?  Or if that is not convenient, I think we can safely merge this branch into master.

maxm17:
@andy-beer It may be better if you could check that it is working correctly, as I don't know what is expected. The test is in master: unit_tests/grid/test_faults.py::test_find_faults_create_organizing_objects_where_needed_true. You will need to uncomment the act and assertion lines

maxm17 (approved):

PR #410, opened 2022-02-01
Vdb numpy bug

maxm17:
Could someone please check my tests as I've not worked with byte arrays before

### DIFF
--- a/resqpy/olio/vdb.py
+++ b/resqpy/olio/vdb.py
@@ -150,8 +150,8 @@ class Header():
         fp.seek(place)
         block = fp.read(22)
         #      log.debug(f'Header block at place {place} returned {len(block)} bytes')
-        self.previous, self.next, c, self.bytes_per_item, self.number_of_items, self.first_fragment, self.max_items =  \
-           unpack('=IIcBIII', block)
+        self.previous, self.next, c, self.bytes_per_item, self.number_of_items, self.first_fragment, self.max_items = \
+            unpack('=IIcBIII', block)
         try:
             self.item_type = c.decode()
         except Exception:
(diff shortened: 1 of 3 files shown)


cflynn3 (approved):
This looks good to me, but let's wait for @andy-beer to review it in case there's anything I've missed.

tests/unit_tests/olio/test_vdb.py
```diff
@@ -0,0 +1,24 @@
+import pytest
+
+import resqpy.olio.vdb as vdb
+import numpy as np
+
+
+@pytest.mark.parametrize("byte_array, expected_array", [(b" \x01\x00\x00\xae\x01\x00\x00", np.array([288, 430])),
```
andy-beer, line 7: I suggest replacing the leading space in the first byte array with the hexadecimal form, \x20 I believe.

resqpy/olio/vdb.py
```diff
@@ -221,8 +221,8 @@ def __init__(self, fp, place, item_type, count, max_count):
             self.a = np.empty((count, 2), dtype = int)
             self.a[:, :] = nda[:, :]
         else:
-            self.a = np.empty((count,), dtype = dtype)
-            self.a.data = fp.read(count * byte_size)  # todo: check C ordering; endianess etc.
+            b = fp.read(count * byte_size)
+            self.a = np.frombuffer(buffer = b, dtype = dtype, count = count)
```
andy-beer, line 225: Does frombuffer() have an explicit byte size per element, or is it deduced from count (and the size of buffer), or does dtype determine the size per element?  Vdb is using 32bit floats and ints.

andy-beer (commented):
Looks good.  I'd like to see one or two of the bp notebooks tested with the new code.

tests/unit_tests/olio/test_vdb.py
maxm17: Added

resqpy/olio/vdb.py
maxm17: It is deduced from count [https://numpy.org/doc/stable/reference/generated/numpy.frombuffer.html](url)

maxm17:
@andy-beer Are there any specific notebooks I should use?

maxm17:
Tested on the wren data vdb files, which produces the same output

Issue #416, opened 2022-02-04
Facet info not being set for saturation arrays during import_vdb_ensemble


maxm17:
facet_type is None, should be 'what'.
facet is None, should be 'oil', 'water', or 'gas'.

PR #419, opened 2022-02-07
setting facet info when importing vdb properties

andy-beer:
Max, I think these changes can go into your test_import_vdb_ensemble branch – though I haven't tested them!


andy-beer:
Resolves issue #416

PR #422, opened 2022-02-07
crs_root changes

maxm17:


resqpy/grid/grid.py
```diff
@@ -505,7 +505,7 @@ def off_handed(self):
         """Returns False if IJK and xyz have same handedness, True if they differ."""
 
         ijk_right_handed = self.extract_grid_is_right_handed()
-        assert rqet.find_tag_text(self.crs_root, 'ProjectedAxisOrder').lower() == 'easting northing'
+        assert self.crs.axis_order.lower() == 'easting northing'
```
andy-beer, line 508: I'd like to see a defensive call to the self.set_crs() method in each function that uses self.crs.  This is to cater for ad hoc construction of a new Grid object, when the crs is not necessarily established by the init method.

resqpy/grid/grid.py
```diff
@@ -544,21 +544,19 @@ def xy_units(self):
         :meta common:
         """
 
-        crs_root = self.extract_crs_root()
-        if crs_root is None:
+        if self.crs is None:
             return None
-        return rqet.find_tag(crs_root, 'ProjectedUom').text
+        return self.crs.xy_units
```
andy-beer, line 549: please add self.set_crs(), as above

resqpy/grid/grid.py
```diff
@@ -544,21 +544,19 @@ def xy_units(self):
         :meta common:
         """
 
-        crs_root = self.extract_crs_root()
-        if crs_root is None:
+        if self.crs is None:
```
andy-beer, line 547: please add self.set_crs() before this line, as above

resqpy/grid/grid.py
```diff
@@ -544,21 +544,19 @@ def xy_units(self):
...
 
     def z_units(self):
         """Returns the vertical (z) units of measure of the coordinate reference system for the grid.
 
         :meta common:
         """
 
-        crs_root = self.extract_crs_root()
-        if crs_root is None:
+        if self.crs is None:
```
andy-beer, line 557: please add self.set_crs() before this line, as above

andy-beer (changes requested):
Please add some more calls to set_crs() as indicated, then good to merge.

andy-beer (approved):

PR #433, opened 2022-02-14
Test against older versions of dependencies

### LINKED ISSUE #432: Test against lowest-supported dependencies
At present, our CI suite tests using the latest available 3rd party libraries. It would be useful to also include a test run against the _lowest_ 3rd party libraries that we claim to support , to validate that resqpy still functions as expected.

The versions we claim to support are specified in `pyproject.toml` (e.g. `numpy = "^1.19"` ), and the exact versions used in current CI tests are in `poetry.lock` (e.g. `numpy = ">=1.14.5" `).

I would recommend we follow the [NEP 29](https://numpy.org/neps/nep-0029-deprecation_policy.html) standard for scientific computing libs, which essentially to support 2 year's worth of minor versions of numpy / pandas etc.

```
On Dec 26, 2021 drop support for Python 3.7 (initially released on Jun 27, 2018)
On Jun 21, 2022 drop support for NumPy 1.19 (initially released on Jun 20, 2020)
On Apr 14, 2023 drop support for Python 3.8 (initially released on Oct 14, 2019)
```

connortann:
Closes #432 

Adds a CI job that runs the unit tests against some older versions of our 3rd party libraries. 

Steps in the job:

1. Installs the usual dependencies, using the reusable "prepare poetry" action implemented in #431
2. Downgrades numpy to 1.19 and pandas to 1.1 (the lowest versions we claim to support in our `pyproject.toml` metadata
3. Runs pytest

I note that for some even older versions of pandas, the "downgrade" step takes ages (10+ minutes) then fails, as pip tries to build pandas from source rather than use a pre-built wheel. However, for the minimum versions we have now, the installation step works reasonably smoothly (~30 seconds).

### DIFF
--- a/.github/workflows/ci-tests.yml
+++ b/.github/workflows/ci-tests.yml
@@ -1,6 +1,6 @@
 ---
 
-name: Python CI
+name: Tests
 
 on:
   push:
@@ -63,8 +63,7 @@ jobs:
           python-version: ${{ matrix.python-version }}
 
       - name: Run pytest
-        run: |
-          poetry run pytest --cov=resqpy --junitxml=pytest.xml
+        run: poetry run pytest --cov=resqpy --junitxml=pytest.xml
 
       - name: Upload pytest artifacts
         if: ${{ always() }}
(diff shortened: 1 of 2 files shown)


andy-beer (approved):
Looks fine to me at a glance but I haven't tried running the tests

andy-beer:
LGTM

PR #436, opened 2022-02-15
Fixing further publish problems.

cflynn3:


pyproject.toml
```diff
@@ -55,7 +55,16 @@ autoclasstoc = "^1.2"
 [tool.poetry-dynamic-versioning]
 enable = true
 vcs = "git"
-style = "semver"
+style = "pep440"
```
connortann, line 58: To help me understand this PR, could you elaborate on the motivation to the changes to the "style" and "format-jinja" options?

We have previously been using Semantic Versioning, so it would be useful to continue in the same style if possible. I don't know of any differences between "semver" and "pep440" that apply to our use case, as we're just using straightforward versions of the form "X.Y.Z", as described in the contributing guide.

From reading the documentation of poetry-dynamic-versioning, it looks like the `style` option defines a custom format, so I'm not clear why a bespoke format pattern would be necessary for our rather vanilla use case?

.github/workflows/ci-publish.yml
```diff
@@ -85,17 +87,22 @@ jobs:
         with:
           python-version: '3.9'
 
+      - name: Checkout code
+        uses: actions/checkout@v2
+        with:
+          fetch-depth: 0
```
connortann, line 93: Should this step need a working copy of the source code, and git histories? The publish step should use the built wheel, so I don't think it should need the source code or the tags (the fetch-depth parameter).

I had rather liked the simplicity of the previous step that did not checkout the code; it simply grabbed the wheel artefact and published to PyPI with twine. Would that still be appropriate here?

.github/workflows/ci-publish.yml
```diff
@@ -53,6 +53,8 @@ jobs:
 
       - name: Checkout code
         uses: actions/checkout@v2
+        with:
+          fetch-depth: 0
```
connortann, line 57: Does this step really need the git history?

.github/workflows/ci-publish.yml
cflynn3: To explain why this step is necessary: the Poetry Publish step looks at the version in pyproject.toml and finds files related to the version being published, then publishes only those files. If it can't find files related to the current version it doesn't publish anything. The problem was that Poetry was unable to get the 'current' version without the source code for the versioning plugin to calculate it.

With regards to the overall question of whether we should revert to Twine, I see the arguments as follows: we now have a working system that keeps the dependencies, runtime, build system and publish all within one ecosystem, which I consider to be a preferable in the long term. If we do revert back to Twine, that would have to be re-tested, however it would require less steps in the publish action, which would make these steps easier to follow and maintain.

I don't have strong feelings either way, and will revert to Twine if that is considered the better option.

connortann: Makes sense, thanks for the explanation. I agree with your points about keeping it in the same ecosystem. Would suggest adding a comment to this step just summarising the above, as I think it's a bit of a subtle point

pyproject.toml
cflynn3: I thought that moving to the PEP440 standard would be preferable, and I find it to be slightly clearer. To clarify, the only difference we will see is when developing software, the semver system produces versions like '1.7.1.post12xxx', wheres PEP440 gives '1.7.1.dev12'. Both versions are exactly the same for tagged releases, and I can revert this change if preferred.

connortann:
May I also suggest rebasing these commits on top of master and squashing away any debugging commits? I think this would be good practice prior to settings PRs as ready to merge - it reduces likelihood of messy git history entering master (as happened recently), and also mitigates the need for squashing upon merges.

pyproject.toml
connortann: Happy with that, agreed it doesn't make much difference.
What about the "format-jinja" option, is that necessary?

cflynn3: It clarifies the behaviour of the versioning library for PEP440, and provides us an easy way to change this behaviour if we desire to in the future.

connortann (approved):
LGTM

PR #440, opened 2022-02-15
Add support for python 10

connortann:
Exploring whether resqpy can support Python 10. Confirmed on CI, it all looks good!

As per discussion on #432 , we might want to consider dropping support for python 3.7 . On


andy-beer (approved):
Good news that resqpy works with python 3.10

andy-beer:
LGTM

PR #444, opened 2022-02-18
deletion of unused olio modules and functions; removal of deprecated functionality

andy-beer:
This change:

-  deletes olio modules that were unused by resqpy
- removes deprecated functionality

We can also remove the beta tag as part of this breaking change, in preparation for a major release tag (v2.0.0)


andy-beer:
Note that this change is a **breaking change**, likely to require changes to application code.

resqpy/grid/_grid.py
```diff
@@ -506,7 +492,10 @@ def off_handed(self):
         """Returns False if IJK and xyz have same handedness, True if they differ."""
 
         ijk_right_handed = self.extract_grid_is_right_handed()
-        assert rqet.find_tag_text(self.crs_root, 'ProjectedAxisOrder').lower() == 'easting northing'
+        if self.crs is None:
+            assert self.crs_uuid is not None
```
cflynn3, line 496: A general comment, but I much prefer to see a message after assertions like this to aid debugging if the situation occurs.

resqpy/olio/uuid.py
```diff
@@ -194,3 +197,16 @@ def version_string(uuid_obj):
...
+    """Returns boolean indicating whether uuid_obj seems to be a uuid."""
+
+    if isinstance(uuid_obj, uuid.UUID):
+        return True
+    if not uuid_obj or not isinstance(uuid_obj, str):
+        return False
+    if uuid_obj[0] == '_':
+        return len(uuid_obj) == 37
+    return len(uuid_obj) == 36
+    # could also check for hyphens in correct places, and hexadecimals only
```
cflynn3, line 212: Agreed, I'm sure there must be some code somewhere that can validate a UUID fully.

cflynn3 (approved):
I've made a couple of general comments here, but I'm happy for this to be merged as is if you don't wish to act on them now.

resqpy/olio/uuid.py
andy-beer: Yes, though I'm happy to let this go for now, as it is unlikely that something like-a-valid-UUID-but-not-quite will show up.

PR #479, opened 2022-04-20
Delaunay triangulation scipy

maxm17:


resqpy/surface/_surface.py
```diff
@@ -398,7 +398,7 @@ def set_from_point_set(self,
         log.debug('number of points going into dt: ' + str(len(p_xy_e)))
         success = False
         try:
-            t = triangulate.dt(p_xy_e[:, :2], container_size_factor = convexity_parameter)
+            t = triangulate.dt(p_xy_e[:, :2], container_size_factor = convexity_parameter, algorithm = "simple")
```
andy-beer, line 401: We will probably want to switch to using scipy here and elsewhere.  Let's merge in as is and we can make that change later.

andy-beer (approved):
Good work, thanks Max

PR #491, opened 2022-05-03
optimisation implemented and unit tests added

maxm17:


maxm17:
image

resqpy/grid_surface.py
```diff
@@ -1494,6 +1497,351 @@ def find_faces_to_represent_surface_regular(grid,
...
+
+def find_faces_to_represent_surface_regular_optimised(
+    grid,
+    surface,
+    name,
+    title = None,
+    centres = None,
+    progress_fn = None,
+    consistent_side = False,
+    return_normal_vectors = False,
```
andy-beer, line 1612: Please can we make this compatible with the latest _regular() version?  (Will require manual merging of the offset related code.)

resqpy/olio/vector_utilities.py
```diff
@@ -523,13 +524,85 @@ def points_in_triangles(p, t, da, projection = 'xy', edged = False):
         return np.all(cwtd < 0.0, axis = 2)
 
 
+@njit
+def point_in_polygon(x, y, polygon):
+    """Calculates if a point in within a polygon in 2D.
+
```
andy-beer, line 530: Does this function require the polygon to be convex? (It doesn't matter for our triangles but would be good to make clear in docstring for other uses.)

andy-beer:
@connortann Could you please take a look, especially at the poetry & toml changes.

pyproject.toml
```diff
@@ -48,6 +48,7 @@ h5py = "^3.2"
 lxml = "^4.6"
 lasio = "^0.29"
 scipy = "^1.7.3"
+numba = "^0.55.0"
```
connortann, line 51: This states that we require the very latest version of numba, minor version `0.55` and nothing else. Can we not have a more flexible constraint, e.g. `numba = "^0.50` ?

This will allow compatibility with recent minor versions and will also allow us to keep up with future non-breaking updates?

resqpy/grid_surface.py
maxm17: Yeah I've just added that

resqpy/olio/vector_utilities.py
maxm17: I think so but I could add some tests for that?

pyproject.toml
maxm17: That is the minimum requirement, otherwise it doesn't satisfy the Python version constraints

connortann: Could you clarify, which python version constraints would be incompatible with e.g. numba `v0.50` ?

resqpy/olio/vector_utilities.py
andy-beer: Yes, I think it would be good to know, so a unit test directly calling this function for a few different shapes and points would be good.  By the way, we also have similar functionality in the olio/point_inclusion module.

andy-beer (approved):

pyproject.toml
connortann: Thanks for the discussion Max and for testing the other versions! Looks like we should go for `^0.52`.

connortann (changes requested):
As per discsussion above, I think the version constraint needs to be tweaked

PR #502, opened 2022-05-23
Adding in a message to make clear that 32-bit Python no longer works with RESQPY due to the HDF5 dependency.

cflynn3:


andy-beer (approved):
The RESQML standard requires 64 bit representation of data, so this constraint is in line with that requirement.

Good to highlight it the way you have.

PR #505, opened 2022-05-30
Multiprocessing Module Added

maxm17:


resqpy/multiprocessing/multiprocessing.py
```diff
@@ -0,0 +1,112 @@
...
+    else:
+        model_recombined = new_model(epc_file = str(epc_file))
+
+    log.info("Creating the recombined epc file.")
+    for i, epc in enumerate(epc_list):
+        if epc is None:
+            continue
+        while True:
+            try:
+                model = Model(epc_file = epc)
```
andy-beer, line 94: Suggest we put a pause of a second or a few seconds in this while loop, to save thrashing the file system.

resqpy/multiprocessing/multiprocessing.py
```diff
@@ -0,0 +1,112 @@
...
+            where the combined epc will be saved.
+        cluster (LocalCluster/JobQueueCluster): a LocalCluster is a Dask cluster on a
+            local machine. If using a job queing system, a JobQueueCluster can be used
+            such as an SGECluster, SLURMCluster, PBSCluster, LSFCluster etc.
+        consolidate (bool): if True and an equivalent part already exists in
+            a model, it is not duplicated and the uuids are noted as equivalent.
+
+    Returns:
+        success_list (List[bool]): A boolean list of successful function calls.
+    """
```
andy-beer, line 63: Could be good to add a note to the docstring, saying that dask is used as the backend and needs to have been installed in the python environment (for people who don't read the tutorial!)

andy-beer (changes requested):
Max, this is all good – a couple of small changes suggested. I'd like the pause at the end of the while loop (especially) to be added before we merge.

resqpy/multiprocessing/multiprocessing.py
maxm17: I can add the link to the tutorial once it is finished too

PR #526, opened 2022-07-01
Drop python 3.7

connortann:
[CVE-2021-34141](https://nvd.nist.gov/vuln/detail/cve-2021-34141) has been given for our currently pinned version of numpy (1.21.6), which is resolved in numpy 1.22. Update numpy requires us to drop support for python 3.7.

The CVE seems to be of questionable importance; the maintainers describe it as "[completely harmless](https://github.com/numpy/numpy/issues/18993)". 

But in any case, it is time for us to drop support for 3.7 according to the [NEP29 deprecation policy](https://numpy.org/neps/nep-0029-deprecation_policy.html):
- Support all minor versions of Python released in the prior 42 months
- Support all minor versions of NumPy/Pandas released in the prior 24 months

As this is a potentially breaking change, I suggest incrementing the semver major version on our next release.

### DIFF
--- a/poetry.lock
+++ b/poetry.lock
@@ -47,7 +47,7 @@ tests = ["pytest", "pytest-cov", "parametrize-from-file", "voluptuous", "coveral
 
 [[package]]
 name = "babel"
-version = "2.10.1"
+version = "2.10.3"
 description = "Internationalization utilities"
 category = "dev"
 optional = false
@@ -58,7 +58,7 @@ pytz = ">=2015.7"
 
 [[package]]
 name = "certifi"
-version = "2022.5.18.1"
+version = "2022.6.15"
 description = "Python package for providing Mozilla's CA Bundle."
 category = "dev"
 optional = false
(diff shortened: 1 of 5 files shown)


poetry.lock
```diff
@@ -303,11 +284,11 @@ numpy = ">=1.18,<1.23"
 
 [[package]]
 name = "numpy"
-version = "1.21.6"
+version = "1.22.4"
 description = "NumPy is the fundamental package for array computing with Python."
 category = "main"
 optional = false
-python-versions = ">=3.7,<3.11"
+python-versions = ">=3.8"
```
cflynn3, line 291: It's probably worth keeping in the maximum version as well, in case we run into more issues with changes breaking our code or (more likely) one of our dependencies.

connortann: Agreed. Python 3.11 is not released yet, but as soon as it is we can pretty easily confirm it works on CI and then support it.

resqpy/grid_surface.py
```diff
@@ -1508,7 +1508,7 @@ def find_faces_to_represent_surface_regular(grid,
 
 
 @njit
-def where_true(data: np.ndarray) -> np.ndarray:
+def where_true(data: np.ndarray):
```
andy-beer, line 1511: Is this an intentional change? Is it required? The just-in-time compilation works best with full type information, as I understand it. (Admittedly in this case it probably doesn't make any difference – I am curious about the change though.)

andy-beer (approved):
I'm happy with dropping support for python 3.7 and can update the major indicator on next release.  I'm curious about the jit related change – if anyone could respond to my review comment I'd appreciate it.

resqpy/grid_surface.py
connortann: This change seemed to be required to get the `mypy` type checker job to pass. I think the type information provided by numpy was updated in the more recent versions, which conflicted with the given return value here.

Given than `np.where` is [natively supported](https://numba.pydata.org/numba-doc/dev/reference/numpysupported.html#other-functions) by numba, I think it should be fine to include within a jitted function even without a return type hint.

connortann: By all means we could re-add the type information in a future change, if it can be done in a way that passes type checking

PR #529, opened 2022-07-05
Wellspec changes

maxm17:


tests/unit_tests/olio/test_wellspec_keywords.py
```diff
@@ -47,172 +46,98 @@ def test_check_value_known_keyword(keyword, value, boolean_return_expected):
...
     assert well_dict == {}
+
+
+def test_get_well_pointers(wellspec_file_two_wells):
+    # Act
+    well_pointers = wk.get_well_pointers(wellspec_file_two_wells)
+
+    # Assert
+    assert len(well_pointers) == 2
+    assert well_pointers["TEST_WELL1"] == 21
```
andy-beer, line 121: These pointer values could potentially be system dependent.  I'm okay for the test to go in like this – if it runs on my Mac and on the hpc then we are probably okay, though it would be good to execute the tests on a Windows laptop and on Azure at some point.  Don't worry about it just now though.

andy-beer (approved):

PR #546, opened 2022-07-26
Polygons Points Memory Reduced

maxm17:


andy-beer (approved):
This looks fine Max.  There were a few flake8 errors relating to docstrings in the vector_utilities module – please correct those then we can merge.

PR #551, opened 2022-07-28
Column Property Wrapper Function

maxm17:


resqpy/multiprocessing/wrappers/mesh_mp.py
```diff
@@ -0,0 +1,125 @@
...
+    if g_crs_uuid is not None:
+        model.copy_uuid_from_other_model(g_model, g_crs_uuid)
+        uuid_list.append(g_crs_uuid)
+    model.copy_uuid_from_other_model(g_model, uuid = grid_uuid)
+
+    for prop_uuid in prop_uuids:
+        model.copy_uuid_from_other_model(g_model, uuid = prop_uuid)
+
+    grid = RegularGrid(parent_model = model, uuid = grid_uuid)
+
```
andy-beer, line 64: I think we should probably add the grid uuid to the uuid list.

andy-beer (approved):
This all looks good Max.  I'm going to merge as is.  We might need to add the grid uuid to the uuid list, so that we end up with a copy of it when creating a new epc (though not strictly required by the RESQML standard, parts of resqpy will be unhappy if the grid is missing).

PR #556, opened 2022-08-02
Add generation of a boolean flange property to TriangulatedSets when extending with flange

emmanesbit:


resqpy/multiprocessing/wrappers/grid_surface_mp.py
```diff
@@ -158,19 +172,31 @@ def find_faces_to_represent_surface_regular_wrapper(
...
         retriangulated = True
     if not bu.matching_uuids(surface.uuid, surface_uuid):
         surface.write_hdf5()
         surface.create_xml()
         surface_uuid = surface.uuid
+    if flange_bool is not None:
+        flange_bool_property = Property.from_array(parent_model = model,
+                                                   cached_array = flange_bool,
+                                                   source_info = 'flange bool array',
+                                                   keyword = 'flange bool property',
```
andy-beer, line 192: let's drop 'property' from the title here; 'flange bool' or simply 'flange' is better I feel

emmanesbit:
@andy-beer i've updated the property variable as suggested, and added an additional unit test to check for the bool property. Could you please rerun yapf and merge when happy? :)

PR #558, opened 2022-08-03
Blocked well dateformat

poc11:
When creating a BlockedWell object from a wellspec file, the user must specify whether the usa_date_format is True (MM/DD/YYYY) or False (DD/MM/YYYY). Currently the usa_date_format variable defaults to False and there's no way to specify otherwise when creating a BlockedWell object. This PR adds the usa_date_format arg to the BlockedWell constructor as well as several supporting functions within resqpy/well/well_object_funcs.py.


andy-beer:
Looks like yapf does not replace tabs with spaces, which is what needs to be done here. (My code editor is set to do that all the time so I haven't seen this error before.)

PR #559, opened 2022-08-03
Add flange bool property to grid connection set

emmanesbit:


resqpy/grid_surface.py
```diff
@@ -1923,6 +1931,18 @@ def find_faces_to_represent_surface_regular_optimised(
         # log.debug(f'gcs count: {gcs.count}; all offsets shape: {all_offsets.shape}')
         assert all_offsets.shape == (gcs.count,)
 
+    if return_flange_bool:
+        all_flange = np.full((grid.nk, grid.nj, grid.ni - 1), -1, dtype = bool)
```
andy-beer, line 1935: I think this line (all_flange = np.full(...)) should be deleted.  Or am I missing something here?

andy-beer (changes requested):
A couple of lines to be deleted I believe, otherwise looks good.

PR #563, opened 2022-08-11
Small Changes To intersect_numba

maxm17:


andy-beer (approved):
This looks good to merge to me.  At a later date we might consider moving to a 'sparse' representation for the arrays (faces, offsets, normals etc.) but it would means a lot code changes elsewhere, so let's park the idea for now.

PR #567, opened 2022-08-19
Update gcs to write tmult values to nexus fault file

emmanesbit:
This update allows a trans_mult_uuid to be provided when using the 'write_simulator' method. If a uuid is provided, and this property exists on the gcs these values will be used as transmissibility multiplier values on the output file. If not provided, the default of 1.0 is used.

### DIFF
--- a/resqpy/fault/_grid_connection_set.py
+++ b/resqpy/fault/_grid_connection_set.py
@@ -1301,7 +1301,7 @@ class GridConnectionSet(BaseResqpy):
                     both = np.empty((cell_index_pairs.shape[0], 6), dtype = int)  # axis, polarity, k, j, i, tmult
                     both[:, :2] = face_index_pairs[:, side, :]  # axis, polarity
                     both[:, 2:-1] = cell_index_pairs[:, side, :]  # k, j, i
-                    both[:, -1:] = feat_mult_array.reshape(-1, 1)
+                    both[:, -1] = feat_mult_array.flatten()
                     df = pd.DataFrame(both, columns = ['axis', 'polarity', 'k', 'j', 'i', 'tmult'])
                     df = df.sort_values(by = ['axis', 'polarity', 'j', 'i', 'k', 'tmult'])
                     both_sorted = np.empty(both.shape, dtype = int)
@@ -1309,7 +1309,7 @@ class GridConnectionSet(BaseResqpy):
                     cell_indices = both_sorted[:, 2:-1]
                     face_indices = np.empty((both_sorted.shape[0], 2), dtype = int)
                     face_indices[:, :] = both_sorted[:, :2]
-                    tmult_values = both_sorted[:, -1:]
+                    tmult_values = both_sorted[:, -1]
                     del both_sorted
                     del both
                     del df
@@ -1318,7 +1318,7 @@ class GridConnectionSet(BaseResqpy):
                     for row in range(cell_indices.shape[0]):
                         kp, jp, ip = cell_indices[row]
                         axis_p, polarity_p = face_indices[row]
-                        tmult = tmult_values[row][0]
+                        tmult = tmult_values[row]
                         if k is not None:
                             if axis_p != axis or polarity_p != polarity or ip != i or jp != j or kp != k2 + 1:
                                 write_row(self, fp, feature_name, i, j, k, k2, axis, polarity, tmult)
--- a/resqpy/fault/_grid_connection_set.py
+++ b/resqpy/fault/_grid_connection_set.py
@@ -1238,10 +1238,10 @@ class GridConnectionSet(BaseResqpy):
 
     def write_simulator(self,
                         filename, 
-                        mode = 'w', 
+                        mode = 'w',
                         simulator = 'nexus', 
-                        include_both_sides = False, 
-                        use_minus = False, 
+                        include_both_sides = False,
+                        use_minus = False,
                         trans_mult_uuid = None):
         """Creates a Nexus include file holding MULT keywords and data. trans_mult_uuid (optional) is the uuid of a property on the gcs containing transmissibility multiplier values. If not provided values of 1.0 will be used."""
         
@@ -1250,7 +1250,7 @@ class GridConnectionSet(BaseResqpy):
                 trans_mult_uuid)), f'trans_mult_uuid provided is not part of collection {trans_mult_uuid}'
             tmult_array = self.property_collection.cached_part_array_ref(self.model.part_for_uuid(trans_mult_uuid))
             assert tmult_array is not None
-        else: 
+        else:
             tmult_array = None
 
         assert simulator == 'nexus'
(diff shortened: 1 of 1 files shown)


resqpy/fault/_grid_connection_set.py
```diff
@@ -1236,8 +1236,21 @@ def create_xml(self,
...
+    def write_simulator(self,
+                        filename,
+                        mode = 'w',
+                        simulator = 'nexus',
+                        include_both_sides = False,
+                        use_minus = False,
+                        trans_mult_uuid = None):
+        """Creates a Nexus include file holding MULT keywords and data. trans_mult_uuid (optional) is the uuid of a property on the gcs containing transmissibility multiplier values. If not provided values of 1.0 will be used."""
+        if trans_mult_uuid is not None:
+            assert self.property_collection.part_in_collection(self.model.part_for_uuid(
```
andy-beer, line 1248: Best to call the self.extract_property_collection() method before this assertion, to ensure that the property_collection is populated.

resqpy/fault/_grid_connection_set.py
```diff
@@ -1258,11 +1271,11 @@ def write_nexus_header_lines(fp, axis, polarity, fault_name, grid_name = 'ROOT')
                 tm.log_nexus_tm('warning')
             active_nexus_head = (axis, polarity, fault_name, grid_name)
 
-        def write_row(gcs, fp, name, i, j, k1, k2, axis, polarity):
+        def write_row(gcs, fp, name, i, j, k1, k2, axis, polarity, tmult):
             nonlocal row_count
             write_nexus_header_lines(fp, axis, polarity, name)
-            fp.write('\t{0:1d}\t{1:1d}\t{2:1d}\t{3:1d}\t{4:1d}\t{5:1d}\t1.0\n'.format(
-                i + 1, i + 1, j + 1, j + 1, k1 + 1, k2 + 1))
+            fp.write('\t{0:1d}\t{1:1d}\t{2:1d}\t{3:1d}\t{4:1d}\t{5:1d}\t{6:.2f}\n'.format(
```
andy-beer, line 1277: I think we need at least 3 decimal places for multipliers: Believe it or not, REs can want to set a multiplier of 0.001 and regard it as different to zero.  I suggest :.4f for the tmult format (instead of :.2f), to be on the safe side.

resqpy/fault/_grid_connection_set.py
```diff
@@ -1278,17 +1291,24 @@ def write_row(gcs, fp, name, i, j, k1, k2, axis, polarity):
...
                 feature_name = self.feature_list[feature_index][2].split()[0].upper()
                 cell_index_pairs, face_index_pairs = self.list_of_cell_face_pairs_for_feature_index(feature_index)
+                if tmult_array is not None:
+                    feature_mask = np.where(self.feature_indices == feature_index, 1, 0)
+                    feat_mult_array = np.extract(feature_mask, tmult_array)
+                else:
+                    feat_mult_array = np.ones(shape = (cell_index_pairs.shape[0],), dtype = float)
                 for side in sides:
-                    both = np.empty((cell_index_pairs.shape[0], 5), dtype = int)
+                    both = np.empty((cell_index_pairs.shape[0], 6), dtype = int)
```
andy-beer, line 1300: Let's add a trailing comment on the np.empty() line, indicating what the 6 elements will be, something like:
# axis, polarity, k, j, i, tmult

andy-beer: Ha, I used a hash in my comment and it went into markdown mode!

resqpy/fault/_grid_connection_set.py
```diff
@@ -1278,17 +1291,24 @@ def write_row(gcs, fp, name, i, j, k1, k2, axis, polarity):
...
                     both[:, :2] = face_index_pairs[:, side, :]  # axis, polarity
-                    both[:, 2:] = cell_index_pairs[:, side, :]  # k, j, i
-                    df = pd.DataFrame(both, columns = ['axis', 'polarity', 'k', 'j', 'i'])
-                    df = df.sort_values(by = ['axis', 'polarity', 'j', 'i', 'k'])
+                    both[:, 2:-1] = cell_index_pairs[:, side, :]  # k, j, i
+                    both[:, -1:] = feat_mult_array.reshape(-1, 1)
```
andy-beer, line 1303: both[:, -1] is slightly cleaner than both[:, -1:] and means the same thing

resqpy/fault/_grid_connection_set.py
```diff
@@ -1278,17 +1291,24 @@ def write_row(gcs, fp, name, i, j, k1, k2, axis, polarity):
...
+                    df = pd.DataFrame(both, columns = ['axis', 'polarity', 'k', 'j', 'i', 'tmult'])
+                    df = df.sort_values(by = ['axis', 'polarity', 'j', 'i', 'k', 'tmult'])
```
andy-beer, line 1305: Not sure sorting by tmult adds anything but harmless enough, especially as there should only be one entry for a given axis, polarity, k, j, i combination.

resqpy/fault/_grid_connection_set.py
```diff
@@ -1278,17 +1291,24 @@ def write_row(gcs, fp, name, i, j, k1, k2, axis, polarity):
...
                     both_sorted = np.empty(both.shape, dtype = int)
                     both_sorted[:] = df
-                    cell_indices = both_sorted[:, 2:]
+                    cell_indices = both_sorted[:, 2:-1]
                     face_indices = np.empty((both_sorted.shape[0], 2), dtype = int)
                     face_indices[:, :] = both_sorted[:, :2]
+                    tmult_values = both_sorted[:, -1:]
```
andy-beer, line 1311: Similar to earlier: cleaner to remove that last colon

resqpy/fault/_grid_connection_set.py
```diff
@@ -1278,17 +1291,24 @@ def write_row(gcs, fp, name, i, j, k1, k2, axis, polarity):
...
                 cell_index_pairs, face_index_pairs = self.list_of_cell_face_pairs_for_feature_index(feature_index)
+                if tmult_array is not None:
+                    feature_mask = np.where(self.feature_indices == feature_index, 1, 0)
+                    feat_mult_array = np.extract(feature_mask, tmult_array)
+                else:
+                    feat_mult_array = np.ones(shape = (cell_index_pairs.shape[0],), dtype = float)
                 for side in sides:
-                    both = np.empty((cell_index_pairs.shape[0], 5), dtype = int)
+                    both = np.empty((cell_index_pairs.shape[0], 6), dtype = int)
                     both[:, :2] = face_index_pairs[:, side, :]  # axis, polarity
```
andy-beer, line 1301: Don't know what I clicked that made this line turn yellow in the review display!

andy-beer (changes requested):
This looks good.  Some minor changes suggested, then good to merge.

resqpy/fault/_grid_connection_set.py
andy-beer: if removing that colon, then the reshape should be replaced with a .flatten() I think

andy-beer (approved):
we can run with this

andy-beer (approved):

PR #568, opened 2022-08-24
BlockedWell wrapper and batch processing functions

maxm17:


resqpy/multiprocessing/wrappers/blocked_well_mp.py
```diff
@@ -0,0 +1,133 @@
...
+            - index (int): the index passed to the function.
+            - success (bool): True if all the BlockedWell objects could be created, False
+              otherwise.
+            - epc_file (str): the epc file path where the objects are stored.
+            - uuid_list (List[UUID/str]): list of UUIDs of relevant objects.
+    """
+    uuid_list = []
+    tmp_dir = Path(f"tmp_dir/{uuid.uuid4()}")
+    tmp_dir.mkdir(parents = True, exist_ok = True)
+    epc_file = f"{tmp_dir}/wrapper.epc"
```
andy-beer, line 50: Better to use os.sep than /
Or os.path.join()
Or Path equivalent and str()

resqpy/multiprocessing/wrappers/blocked_well_mp.py
```diff
@@ -0,0 +1,133 @@
...
+    model = new_model(epc_file = epc_file)
+
+    trajectory_model = Model(trajectory_epc)
+    grid_model = Model(grid_epc)
+    model.copy_uuid_from_other_model(grid_model, uuid = grid_uuid)
+
+    grid = RegularGrid(grid_model, grid_uuid)
```
andy-beer, line 57: Best to copy the grid to model as well as the trajectory

resqpy/multiprocessing/wrappers/blocked_well_mp.py
```diff
@@ -0,0 +1,133 @@
...
+
+    success = True
+    for trajectory_uuid in trajectory_uuids:
+        trajectory = Trajectory(
+            trajectory_model,
```
andy-beer, line 62: Better to open the copy of Trajectory that is in model

andy-beer (commented):
This looks good Max.
Please see my comments about copying grid as well as trajectory to the new model, and opening them from there.
Also, could we get some test coverage in at some point please.  (We can merge this without for now and I can test manually on some real data).

resqpy/multiprocessing/wrappers/blocked_well_mp.py
maxm17: Yeah I can change that

resqpy/multiprocessing/wrappers/blocked_well_mp.py
maxm17: No problem, I'll sort that

andy-beer:
Max, Have you checked in the latest blocked_well_mp.py?  It is still showing as the old file path construction for me.

PR #571, opened 2022-08-30
Wrapper function epc paths updated

maxm17:


resqpy/multiprocessing/wrappers/blocked_well_mp.py
```diff
@@ -47,7 +47,7 @@ def blocked_well_from_trajectory_wrapper(
     uuid_list = []
     tmp_dir = Path("tmp_dir") / f"{uuid.uuid4()}"
     tmp_dir.mkdir(parents = True, exist_ok = True)
-    epc_file = f"{tmp_dir}/wrapper.epc"
+    epc_file = str(tmp_dir / "wrapper.epc")
```
andy-beer, line 50: does / in a string expression use os.sep?

maxm17: With pathlib you can use "/" to join paths and then it is converting to a string representation

andy-beer (approved):

PR #576, opened 2022-09-02
New points_in_triangles_aligned function added

maxm17:


resqpy/olio/vector_utilities.py
```diff
@@ -862,6 +862,62 @@ def points_in_triangles_aligned(nx: int, ny: int, dx: float, dy: float, triangle
...
 
+@njit
+def triangle_box(tri):
+    x_values = tri[:, 0]
+    y_values = tri[:, 1]
+
+    def min_max(x):
+        maximum = x[0]
+        minimum = x[0]
+        for i in x[1:]:
```
andy-beer, line 873: We know there are exactly three rows in the tri array, so better to have two if blocks and avoid the overhead of the for loop I think.

resqpy/olio/vector_utilities.py
```diff
@@ -862,6 +862,62 @@ def points_in_triangles_aligned(nx: int, ny: int, dx: float, dy: float, triangle
...
+    for triangle_num in range(len(triangles)):
+        triangle = triangles[triangle_num]
+        box_min_max = triangle_box(triangle)
+        grid_x = np.arange(dx/2, dx/2 + dx * nx, dx)
+        grid_y = np.arange(dy/2, dy/2 + dy * ny, dy)
+        x_values = grid_x[np.logical_and(grid_x>=box_min_max[0], grid_x<=box_min_max[1])]
+        y_values = grid_y[np.logical_and(grid_y>=box_min_max[2], grid_y<=box_min_max[3])]
+        for x in x_values:
+            ys = []
+            ys.append(np.interp(x, triangle[:2, 0], triangle[:2, 1]))
```
andy-beer, line 909: Does this work as coded here?  Reading the docs for np.interp, I see that it returns the end point of the y range when x is outside the x range (unless the optional left, right args are set).

andy-beer (approved):
This looks good Max.  Could you please check docstrings to strip out normal vectors where they are no longer supported as a property to return.

PR #580, opened 2022-09-05
Add method to blocked well to pull grid properties

emmanesbit:
New method on blockedwell to allow a list of uuids for grid properties to be provided, and added as properties on the blockedwell. Only indexableelement of 'cells' accepted.

### DIFF
--- a/resqpy/well/_blocked_well.py
+++ b/resqpy/well/_blocked_well.py
@@ -3163,10 +3163,10 @@ class BlockedWell(BaseResqpy):
         assert len(self.grid_list) == 1, "Only blocked wells with a single grid can be handled currently"
         grid = self.grid_list[0]
         parts = self.model.parts_list_filtered_by_supporting_uuid(part_list,
-                                                                  grid.uuid) # only those properties on the grid
+                                                                  grid.uuid)  # only those properties on the grid
         gridpc = grid.property_collection
         parts = [part for part in parts if gridpc.indexable_for_part(part) == 'cells'
-                ] # only 'cell' properties are handled
+                ]  # only 'cell' properties are handled
 
         if len(parts) > 0:
             bwpc = rqp.PropertyCollection(support = self)
--- a/resqpy/well/_blocked_well.py
+++ b/resqpy/well/_blocked_well.py
@@ -3172,7 +3172,7 @@ class BlockedWell(BaseResqpy):
             bwpc = rqp.PropertyCollection(support = self)
             if len(gridpc.time_series_uuid_list()) > 0:
                 time_dict = {}
-                for parts in parts:
+                for part in parts:
                     if gridpc.time_series_uuid_for_part(part) in time_dict.keys():
                         time_dict[gridpc.time_series_uuid_for_part(
                             part)] = time_dict[gridpc.time_series_uuid_for_part(part)] + [part]
--- a/resqpy/well/_blocked_well.py
+++ b/resqpy/well/_blocked_well.py
@@ -3159,7 +3159,7 @@ class BlockedWell(BaseResqpy):
         """Add properties to blocked wells from a list of uuids for properties on the supporting grid."""
 
         part_list = [self.model.part_for_uuid(uuid) for uuid in uuid_list]
-    
+
         assert len(self.grid_list) == 1, "Only blocked wells with a single grid can be handled currently"
         grid = self.grid_list[0]
         parts = self.model.parts_list_filtered_by_supporting_uuid(part_list,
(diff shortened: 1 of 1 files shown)


resqpy/well/_blocked_well.py
```diff
@@ -3154,3 +3154,58 @@ def write_hdf5(self, file_name = None, mode = 'a', create_for_trajectory_if_need
...
+    def add_grid_property_to_blocked_well(self, uuid_list):
+        """Add properties to blocked wells from a list of uuids for properties on the supporting grid."""
+
+        part_list = [self.model.part_for_uuid(uuid) for uuid in uuid_list]
+
+        assert len(self.grid_list) == 1, "Only blocked wells with a single grid can be handled currently"
+        grid = self.grid_list[0]
+        parts = self.model.parts_list_filtered_by_supporting_uuid(part_list,
+                                                                  grid.uuid)  # only those properties on the grid
+        gridpc = grid.property_collection
```
andy-beer, line 3167: a bit safer to use:
gridpc = grid.extract_property_collection()

that method will populate the property collection if it has not already been populated

resqpy/well/_blocked_well.py
```diff
@@ -3154,3 +3154,58 @@ def write_hdf5(self, file_name = None, mode = 'a', create_for_trajectory_if_need
...
+
+    def add_grid_property_to_blocked_well(self, uuid_list):
+        """Add properties to blocked wells from a list of uuids for properties on the supporting grid."""
+
+        part_list = [self.model.part_for_uuid(uuid) for uuid in uuid_list]
+
+        assert len(self.grid_list) == 1, "Only blocked wells with a single grid can be handled currently"
+        grid = self.grid_list[0]
+        parts = self.model.parts_list_filtered_by_supporting_uuid(part_list,
+                                                                  grid.uuid)  # only those properties on the grid
```
andy-beer, line 3166: could be good to warn if len(parts) < len(part_list), as it means one or more of the uuids was not for a property on the grid

resqpy/well/_blocked_well.py
```diff
@@ -3154,3 +3154,58 @@ def write_hdf5(self, file_name = None, mode = 'a', create_for_trajectory_if_need
...
+        gridpc = grid.property_collection
+        parts = [part for part in parts if gridpc.indexable_for_part(part) == 'cells'
+                ]  # only 'cell' properties are handled
```
andy-beer, line 3169: maybe also a warning here, if len(parts) is less after this than before, then one or more of the uuids was for a non-cell property

resqpy/well/_blocked_well.py
```diff
@@ -3154,3 +3154,58 @@ def write_hdf5(self, file_name = None, mode = 'a', create_for_trajectory_if_need
...
+
+        if len(parts) > 0:
+            bwpc = rqp.PropertyCollection(support = self)
+            if len(gridpc.time_series_uuid_list()) > 0:
+                time_dict = {}
```
andy-beer, line 3174: this dictionary maps from time series uuid to a list of parts I think – I recommend always adding an in-line comment when a dict first appears saying what it is going to map from and to (ie. keys & values).

resqpy/well/_blocked_well.py
```diff
@@ -3154,3 +3154,58 @@ def write_hdf5(self, file_name = None, mode = 'a', create_for_trajectory_if_need
...
+                for part in parts:
+                    if gridpc.time_series_uuid_for_part(part) in time_dict.keys():
+                        time_dict[gridpc.time_series_uuid_for_part(
+                            part)] = time_dict[gridpc.time_series_uuid_for_part(part)] + [part]
+                    else:
+                        time_dict[gridpc.time_series_uuid_for_part(part)] = [part]
+            else:
+                time_dict = {None: parts}
```
andy-beer, line 3182: Hmm, unless I'm missing something, this coding will skip static properties if there are one or more time varying properties present, which is not ideal. I would suggest building a temp list of parts that you have assigned to any entry in the time_dict as you build it up, then in this final else you can simply set that temp list to the empty list [ ]. After that, outside the if else, you could set the time_dict entry for None to be the list of parts that are not in that temp list.

andy-beer (changes requested):
A few, mostly small, comments to attend to please. Please call if you need further clarification.

resqpy/well/_blocked_well.py
andy-beer: I stand corrected.

andy-beer (approved):

PR #585, opened 2022-09-26
Property xml improvements

andy-beer:
- Preferring "rock permeability" in new xml
- Only applying string lookup table to discrete (not continuous) properties
- Quiet argument when loading or saving a model
- Allowing hyphens in recurrent keywords in Nexus vdb handling

Nexus is a trademark of Halliburton


andy-beer:
@emmanesbit, I've allowed several changes to get bundled into this pull request (sorry). Please focus on the property xml changes for any review.  The vdb import piece has been separately tested and does not need further review.  Similarly for the multiprocessing changes.

PR #593, opened 2022-10-17
WIP: auto generation of local property kinds when abstract kind specified

andy-beer:
This change causes automatic generation of a local property kind when adding a property with one of the 3 main abstract property kinds: 'continuous', 'discrete', 'categorical'. This change is needed for inter-operability with recent versions of Fesapi, and for strict adherence to one of the ancillary xml files in the RESQML v2.0.1 standard.


andy-beer:
Currently WIP as multiprocessing code needs to be checked for inclusion of auto-generated property kinds in return list of uuids to be collated.

andy-beer:
Dependency logic ensures that local property kinds will be setup during multiprocessing merging of worker datasets.

PR #601, opened 2022-11-02
better handling of differing xy and z units

andy-beer:
This change enhances various areas of functionality for the situation where a depth based CRS has z units which differ from the xy units.


emmanesbit (approved):
Looks good to me

PR #602, opened 2022-11-08
Temporary directory path and joblib backend options added

maxm17:


resqpy/multiprocessing/multiprocessing.py
```diff
@@ -56,44 +56,45 @@ def function_multiprocessing(
...
+        This function uses the Dask backend by default to run the given function in
+        parallel, so a Dask cluster must be setup and passed as an argument if Dask is
+        used. Dask will need to be installed in the Python environment because it is not
+        a dependency of the project. More info can be found at 
         https://resqpy.readthedocs.io/en/latest/tutorial/multiprocessing.html
     """
     log.info("multiprocessing function called with %s function, %s entries.", function.__name__, len(kwargs_list))
 
-    tmp_dir = f'tmp_{uuid.uuid4()}'
+    tmp_dir = f'{tmp_dir_path}/tmp_{uuid.uuid4()}'
```
andy-beer, line 83: Probably better to use os.sep instead of '/', or use os.path.join(), to construct the path.

andy-beer (changes requested):
One small change suggested, otherwise this looks good.

PR #604, opened 2022-11-28
converting offsets to consistent (z) units

andy-beer:
This change ensures that gcs offset property values are in a consistent unit of measure – the grid's crs z units – when finding faces to represent a surface and the grid has differing xy & z units.


emmanesbit (approved):
Looks good to me

PR #606, opened 2022-11-29
Upgrading dependency versions as well as GitHub Action versions

cflynn3:


cflynn3:
I've upgraded various dependency versions, GitHub actions, as well as fixed a couple of newly discovered MyPy issues. Unfortunately we can't currently upgrade to allow Python 3.11 at this time as [numba still doesn't supprort it](https://github.com/numba/numba/issues/8304), but when they do allow upgrades of the Python version it should be simpler to do now.

pyproject.toml
```diff
@@ -39,28 +39,29 @@ include = ["data/*.json"]
...
-python = ">= 3.8, < 3.11"
-numpy = "^1.22"
-pandas = "^1.1"
-h5py = "^3.2"
-lxml = "^4.6"
-lasio = "^0.29"
-scipy = "^1.7.3"
-numba = ">=0.52, < 1.0"
-joblib = "^1.1.0"
+python = ">= 3.8.1, < 3.12"
```
andy-beer, line 42: is the < 3.12 okay here, given the current numba limitation?

andy-beer (commented):
Only one small query from me (see comment). Otherwise good to merge. Thanks Chris.

pyproject.toml
cflynn3: Yep good catch, I've changed it back for now.

PR #610, opened 2022-12-02
restoring line that clipped trailing C comments

poc11:
Small fix to resqpy.olio.read_nexus_fault.load_nexus_fault_mult_table function
Previous version had commented out line that partitioned comments out by line.partition('C')
Restoring this line - it's still needed for the rare case where a user is still using C for comments.


poc11:
Closing this in favor of [fault incl file concatenation fix #613](https://github.com/bp/resqpy/pull/613)

PR #618, opened 2022-12-12
improvements to grid coarsening etc

andy-beer:
This change includes support for a time_series_uuid when adding a cached array to a PropertyCollection import list. (Previously the time series could only be identified at the later step of creating xml for the import list properties.)


emmanesbit (approved):
Looks good to me

PR #619, opened 2022-12-12
Grid represented interpretation

emmanesbit:
Updates to allow represented interpretations to be set on grid objects


andy-beer (approved):
Looks good Emma

PR #637, opened 2023-01-09
Draft: Feature/numba cuda polygon

TheGitPuller:
Adds GPU-offloading through CUDA for the calculation of grid faces that intersect surface. New function `find_faces_to_represent_surface_regular_cuda_mgpu` is a wrapper function using multi-GPU offloading to call `find_faces_to_represent_surface_regular_cuda_sgpu` which replicates `find_faces_to_represent_surface_regular_optimised` but on the GPU.

### DIFF
--- a/resqpy/grid_surface.py
+++ b/resqpy/grid_surface.py
@@ -35,6 +35,8 @@ import resqpy.surface as rqs
 import resqpy.well as rqw
 from resqpy.property import Property
 
+import matplotlib.pyplot as plt
+
 compiler_lock = threading.Lock() # Numba compiler is not threadsafe
 
 class GridSkin:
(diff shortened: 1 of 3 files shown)


andy-beer:
After discussion, the current idea is to merge this GPU/cuda functionality (once conflicts are resolved), but not to add cuda to the formal dependencies in the resqpy poetry configuration. This means that anyone wanting to call the GPU based function(s) would have to manually install the correct version of cuda for their GPU hardware.

andy-beer:
After some current regression testing, the changes in this pull request will be manually merged with major release change #655  (branch mandir).

andy-beer:
This pull request has been reclassified as Draft, as the changes have been refactored and manually merged into the mandir branch.

PR #641, opened 2023-01-12
blocked well from wellspec handling time varying properties

andy-beer:


andy-beer:
Unit testing of time varying blocked well properties still needed, yet going to merge with this as is.

PR #645, opened 2023-01-21
Model related uuids refactor

andy-beer:
This performance change introduces a new dictionary attribute to the Model class: uuid_rels_dict, and makes use of it when filtering parts list by related uuid.  This reduces the amount of xml searching needed for such filtering, and reduces the number of times hexadecimal strings representing a uuid need to be converted to an int.


andy-beer:
@maxm17 Max, this is ready to be merged now, I think. Perhaps you could cast your eye over it and also run your example notebook with the latest code, as an extra check and to capture some more timings. Thanks!

PR #655, opened 2023-02-01
major change

andy-beer:
- multiprocessing package renamed multi_processing (**breaking change**)
- grid_surface refactored into package
- deprecated code removed (**breaking change**)


andy-beer:
Marking as WIP until manual merge of changes in open GPU/cuda merge request, which will be incorporated into this change.

resqpy/olio/vector_utilities.py
```diff
@@ -749,32 +748,32 @@ def mesh_points_in_triangle(triangle: np.ndarray,
 def points_in_polygons_parallel(points: np.ndarray, polygons: np.ndarray, points_xlen: int) -> np.ndarray:
```
maxm17, line 748: Could probably remove this function

maxm17 (commented):
Looks good but I would consider making some changes. In general, some public functions/ methods are missing types in their docstrings. I have noticed a lot of the docstrings have been changed to "arguments" and "returns" instead of "Args" and "Returns" which is a hybrid of sphinx and Google style docstrings. That is fine as long as it is consistent and the HTML docs are still built correctly throughout

PR #668, opened 2023-02-22
Changes for faster load of epc files

maxm17:


andy-beer (approved):
Looks okay Max @maxm17 – did you check the timings to ensure it goes faster? (I am wondering how much overhead there is in a try clause, for example.)

maxm17:
@andy-beer Yeah I only managed around a 30% speed up but that saves around a minute to the current workflow I am using

PR #669, opened 2023-02-23
Changes for faster writes to epc files

maxm17:


resqpy/olio/xml_et.py
```diff
@@ -713,14 +696,14 @@ def _escaped_text(text):
...
         colon_attrib_key, pre_colon_attrib = colon_prefixed(key)
         if pre_colon_attrib and pre_colon_attrib not in ns_keys:
             attrib_ns_list.append(pre_colon_attrib)
         if match(key, 'type'):
-            type_attr, type_pre_colon = colon_prefixed(root.attrib[key])
+            type_attr, type_pre_colon = colon_prefixed(val)
             attrib_list.append(colon_attrib_key + '="' + type_attr + '"')
-        elif ct_special and colon_attrib_key == 'PartName' and root.attrib[key].startswith('obj_'):
+        elif ct_special and colon_attrib_key == 'PartName' and val.startswith('obj_'):
             attrib_list.append(colon_attrib_key + '="/' + root.attrib[key] + '"')
```
andy-beer, line 707: I believe this _root.attrib[key]_ can be replaced by _val_ (also a couple of lines below)

resqpy/olio/xml_et.py
```diff
@@ -732,43 +715,29 @@ def _escaped_text(text):
         line += ' xmlns:' + type_pre_colon + '="' + ns[type_pre_colon] + '"'
         ns_keys.append(type_pre_colon)
 
-    for attrib in attrib_list:
-        line += ' ' + attrib
+    line += ' ' + ' '.join(attrib_list)
```
andy-beer, line 718: if attrib_list is empty, this will add a trailing space (probably doesn't matter).

resqpy/olio/xml_et.py
maxm17: Yeah that was a change I made, then reverted back by accident

resqpy/olio/xml_et.py
maxm17: Yeah good spot, I'll change it

PR #701, opened 2023-03-22
Update Coverage of JIT functions

maxm17:
`pytest-cov` doesn't yet support the test coverage of functions using the `njit` decorator. These functions have been manually checked and ignored from the coverage reports

### DIFF
--- a/resqpy/grid_surface/_find_faces.py
+++ b/resqpy/grid_surface/_find_faces.py
@@ -1240,13 +1240,13 @@ def get_boundary(  # type: ignore
     return boundary  # type: ignore
 
 
-@njit
+@njit   # pragma: no cover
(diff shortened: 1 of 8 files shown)


resqpy/grid_surface/_find_faces.py
andy-beer, line 1: Hmm, I'm not so keen on auto-reformatting just for the sake of it. In the old days, programmers respected each others stylistic preferences and the way we worked was, when modifying code that had been authored by someone else, one simply adopted their style. It wasn't difficult. But in this brave new world I suppose I have to move with the times. It's a pity though because 'I much prefer single quotes' to "those double scare quotes", for example.

andy-beer (approved):

andy-beer (approved):

PR #705, opened 2023-04-01
Adding unit tests for Grid functions for finding which cell a point is in

andy-beer:


andy-beer:
Also fixes a bug affecting Grid.test_find_cell_for_point_xy()

PR #720, opened 2023-04-25
Support properties that are indexed per-node in more mesh types

jtai-eq:
Writing a property on an unstructured mesh as support is currently supported with "cells" as indexable element, but not "nodes".  It appears that the only change needed to support "nodes" as indexable element for, e.g., a TetraMesh is in the code that verifies the shape of the input array.  Our proposed code change is similar to how nodes were added as indexable elements to structured meshes in [PR #214](https://github.com/bp/resqpy/pull/214).

We would like to add a test to this PR that covers writing nodes-indexed properties on unstructured, and possibly properties on unstructured grids more generally.  Where do you propose that we add these tests, in tests/unit_tests/unstructured or in tests/unit_tests/property?

### DIFF
--- a/resqpy/property/_collection_get_attributes.py
+++ b/resqpy/property/_collection_get_attributes.py
@@ -499,6 +499,8 @@ def _supporting_shape_gridconnectionset(support, indexable_element):
 def _supporting_shape_other(support, indexable_element):
     if indexable_element is None or indexable_element == 'cells':
         shape_list = [support.cell_count]
+    elif indexable_element == 'nodes':
+        shape_list = [support.node_count]
     elif indexable_element == 'faces per cell':
         support.cache_all_geometry_arrays()
         shape_list = [len(support.faces_per_cell)]


jtai-eq:
Thanks for the feedback.  I added a new unit test file in /property, as you suggested.  For now, it only contains one test, which tests writing and re-reading float-valued properties on a TetraMesh support, with nodes or cells as indexable element.  Ideally this file should contain more tests using different unstructured grids, but it is a start.

andy-beer:
The new unit test module is failing on the format checks. We use _yapf_ to auto-format (and check the formatting) for resqpy – with the yapf options specified in the repo configuration. I think you will need to make the required changes on the fork before the code can be merged.

If you have yapf installed then the following command issued from the top level directory of the forked resqpy should make the required format changes:
`yapf -i tests/unit_tests/property/test_unstructured_grid_properties.py`

PR #725, opened 2023-05-18
Add support for python 3.11

connortann:
Now that lasio has made a new release, resqpy should be able to work with python 3.11 . 

https://github.com/kinverarity1/lasio/issues/572


pyproject.toml
```diff
@@ -39,12 +40,12 @@ include = ["data/*.json"]
...
-python = ">= 3.8.1, < 3.11"
+python = ">= 3.8.1, < 3.12"
 numpy = "^1.23"
 pandas = "^1.4"
 h5py = "^3.7"
 lxml = "^4.9"
-lasio = "^0.30"
+lasio = "^0.31"
 scipy = "^1.9"
 numba = ">=0.56, < 1.0"
```
cflynn3, line 50: Please change this to >= 0.57, as that is the first version to support 3.11 (https://github.com/numba/numba/releases/tag/0.57.0). We could also upgrade the version of numpy now as well if that doesn't break anything else.

connortann: As I understand, we don't need to introduce transitive contraints from 3rd party libraries: numba already declares what python versions are supported, so 0.57 will necessarily be picked up if the python env in 3.11 . 

Numba 0.57 hasn't been out long, only a few weeks. Do we have a reason to drop support for numba 0.56?

connortann: As an example: numba 0.56 is used by our main internal 🌈 python kernel, so dropping support for 0.56 would cause an issue there until the 🌈 build pipeline is fixed.

cflynn3: OK, as long as the libraries resolve in that way then I'm happy to leave the versions as is.

cflynn3 (approved):
Looks good to me.

PR #735, opened 2023-06-27
resampling trianglulated surfaces

emmanesbit:
New method on resqpy.surface.Surface to resample a Surface / triangulated set. Each existing triangle is split into 4 new triangles, with the edges defined by the mid-point of each original triangle edge.

Tests also included.

### DIFF
--- a/resqpy/surface/_surface.py
+++ b/resqpy/surface/_surface.py
@@ -939,7 +939,7 @@ class Surface(rqsb.BaseSurface):
         for i in range(len(rt)):
             tris.extend([[rt[i][0], count1 + i, count3 + i], [rt[i][1], count2 + i, count3 + i],
                          [rt[i][2], count1 + i, count2 + i], [count1 + i, count2 + i, count3 + i]])
-        
+
         if title is None: 
             title = self.citation_title
         resampled = rqs.Surface(self.model, 
--- a/resqpy/surface/_surface.py
+++ b/resqpy/surface/_surface.py
@@ -918,7 +918,7 @@ class Surface(rqsb.BaseSurface):
 
         return ep
 
-    def resampled_surface(self, title=None):
+    def resampled_surface(self, title = None):
         """Creates a new triangulated set which is a resampled version of the current triangulated set. Each existing triangle in the tset is divided equally into 4 new triangles.
            
         arguments:
--- a/resqpy/surface/_surface.py
+++ b/resqpy/surface/_surface.py
@@ -941,7 +941,8 @@ class Surface(rqsb.BaseSurface):
                          [rt[i][2], count1 + i, count2 + i], [count1 + i, count2 + i, count3 + i]])
 
         points_unique, inverse = np.unique(allpoints, axis = 0, return_inverse = True)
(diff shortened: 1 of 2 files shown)


tests/unit_tests/surface/test_surface.py
```diff
@@ -977,3 +978,47 @@ def test_adjust_flange_z(example_model_and_crs):
...
+    assert np.max(rp[:, 0]) == np.max(op[:, 0])
+    assert np.min(rp[:, 1]) == np.min(op[:, 1])
+    assert np.max(rp[:, 1]) == np.max(op[:, 1])
+    assert np.min(rp[:, 2]) == np.min(op[:, 2])
+    assert np.max(rp[:, 2]) == np.max(op[:, 2])
+    assert resampled.crs_uuid == surf.crs_uuid
+    assert resampled.citation_title == surf.citation_title
+    assert resampled_name.citation_title == 'testing'
+    assert resampled.extra_metadata == {'resampled from surface': str(surf.uuid)}
+    assert resampled_name.extra_metadata == {'resampled from surface': str(surf.uuid)}
```
andy-beer, line 1024: Could be nice to sample z values at a few x,y points in the original surface, and then repeat for the resampled surface – they should yield the same z values.

resqpy/surface/_surface.py
```diff
@@ -918,6 +918,38 @@ def axial_edge_crossings(self, axis, value = 0.0):
...
+            tris.extend([[rt[i][0], count1 + i, count3 + i], [rt[i][1], count2 + i, count3 + i],
+                         [rt[i][2], count1 + i, count2 + i], [count1 + i, count2 + i, count3 + i]])
+
+        if title is None:
+            title = self.citation_title
+        resampled = rqs.Surface(self.model,
+                                title = title,
+                                crs_uuid = self.crs_uuid,
+                                extra_metadata = {'resampled from surface': str(self.uuid)})
+        resampled.set_from_triangles_and_points(np.array(tris), allpoints)
```
andy-beer, line 949: Sorry Emma but I don't like this implementation. If I am following it correctly, we will end up with duplicated mid-edge points where an edge is common to two of the original triangles.

Also, this will re-triangulate, which is one approach. My feeling is that it might be better to derive the resampled triangles directly from the original ones, thus preserving tears for example.

andy-beer (approved):

resqpy/surface/_surface.py
```diff
@@ -918,6 +918,45 @@ def axial_edge_crossings(self, axis, value = 0.0):
...
 
+    def resampled_surface(self, title = None):
+        """Creates a new triangulated set which is a resampled version of the current triangulated set. Each existing triangle in the tset is divided equally into 4 new triangles.
+           
+        arguments:
+            title (str): a new title for the output triangulated set, if None the title will have the same title as the input triangulated set
+        
+        returns:
+            resqpy.surface.Surface object, with extra_metadata ('resampled from surface': <uuid>), where uuid is the origin surface uuid
+        """
```
andy-beer, line 929: Worth adding a comment about the edge-based alternative possible implementation, in case we need to optimise at a later date.

andy-beer (approved):

PR #736, opened 2023-06-28
adding support for properties on Polyline, PolylineSet, and PointSet

andy-beer:


emmanesbit (approved):
This all looks good @andy-beer

PR #751, opened 2023-08-25
DRAFT: wellspec with dtop dbot

poc11:
I recently encountered a model that used the DTOP and DBOT nexus keywords to define a well.

`WELLSPEC A_20-1`\
`IW JW L  KH RADW   SKIN   RADB WI PPERF  DTOP      DBOT  `\
`1  1  NA NA 0.2500 0.0100 NA   NA 0.0    6325.0    6330.0 `
`1  1  NA NA 0.2500 0.0200 NA   NA 1.0    8332.0    8370.0`

This creates an issue when trying to derive a BlockedWell from a wellspec. Specifically, when `BlockedWell.derive_from_wellspec` calls `BlockedWell.derive_from_dataframe`, there will not be any information in the dataframe's L (layer) column to populate the `cell_kji0 = BlockedWell.__cell_kji0_from_df(df, i)` variable.

Proposed fix as follows:
`BlockedWell.dtop_dbot_to_layers` function converts the existing `DTOP`/`DBOT`columns in the wellspec dataframe into corresponding k-layers. It does this by:
-loading `corp = grid.corner_points()` 
-iterating through nk and checking if the k,j,i min and max depth is between is in between `DTOP` and `DBOT`
-if cell is in the `DTOP`-DBASE span, a new row is added to the modified dataframe with the corresponding L = k+1
-returns modified dataframe with corresponding L = k+1 values

This function is only called if there are non-null values in the `DTOP`and `DBOT` columns.

Additionally, `resqpy.well.well_utils._derive_from_wellspec_verify_col_list` function has been updated to provide `DTOP` and `DBOT` as defaults (along with `IW`, `JW`, `L`, `ANGLA`, `ANGLV`). This ensures the `DTOP` and `DBOT` columns are checked every time a well is loaded from wellspec.`

### DIFF
--- a/resqpy/well/_blocked_well.py
+++ b/resqpy/well/_blocked_well.py
@@ -575,6 +575,38 @@ class BlockedWell(BaseResqpy):
 
         return self.derive_from_dataframe(df, self.well_name, grid, use_face_centres = True, length_uom = length_uom)
 
+    def dtop_dbot_to_layers(self, df, grid):
+        """Populates the wellspec dataframe with corresponding L when DTOP and DBOT are specified but L is not
+        arguments
+          df (pd.DataFrame): Dataframe of wellspec information
+          grid (grid.Grid object): Grid where well is located
+        returns
+          df_expanded (pd.DataFrame): DataFrame of wellspec information where L is populated with the layers corresponding to the DTOP and DBOT depths
+        """
+        corp = grid.corner_points()
+        nk,nj,ni = corp.shape[:3]
+        df_expanded = []
+
+        for ix,data in df.iterrows():
+            if np.isnan(data['L']):
+                dtop = data['DTOP']
+                dbot = data['DBOT']
+                j,i = int(data['JW']), int(data['IW'])
+                for k in range(nk):
+                    if dbot == 'BOT':
+                        data['L'] = k
+                        df_expanded.append(data.copy())
+                    elif (corp[k,j,i,0,:,:,2].min() > dtop) and (corp[k,j,i,1,:,:,2].max() < dbot):
+                        data['L'] = k
+                        df_expanded.append(data.copy())
+            else:
+                df_expanded.append(data.copy())
+
+        df_expanded = pd.DataFrame(df_expanded).reset_index(drop = True)
+        df_expanded[['IW','JW','L']] = df_expanded[['IW','JW','L']].astype(int)
+
+        return df_expanded
+
     def derive_from_wellspec(self,
                              wellspec_file,
                              well_name,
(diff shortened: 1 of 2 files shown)


poc11:
@andy-beer 
Can you have a look at this pull request? First time i've seen DTOP and DBOT used in a nexus deck but it is a legit Nexus keyword.

Have noticed you avoid caching the full cornerpoint array like i did in the `dtop_dbot_to_layers` function. Curious if you're ok with how I pulled the whole corp in `corp = grid.corner_points()` or if there are any better ways to do that.

resqpy/well/_blocked_well.py
```diff
@@ -575,6 +575,38 @@ def set_for_column(self, well_name, grid, col_ji0, skip_inactive = True, length_
...
+        nk,nj,ni = corp.shape[:3]
+        df_expanded = []
+
+        for ix,data in df.iterrows():
+            if np.isnan(data['L']):
+                dtop = data['DTOP']
+                dbot = data['DBOT']
+                j,i = int(data['JW']), int(data['IW'])
+                for k in range(nk):
+                    if dbot == 'BOT':
```
andy-beer, line 596: does DTOP need to be checked when 'BOT' is in use for DBOT? (I haven't looked in the manual!)

resqpy/well/_blocked_well.py
```diff
@@ -575,6 +575,38 @@ def set_for_column(self, well_name, grid, col_ji0, skip_inactive = True, length_
...
+                        data['L'] = k + 1
+                        df_expanded.append(data.copy())
+                    elif (corp[k,j,i,0,:,:,2].min() >= dtop) and (corp[k,j,i,1,:,:,2].max() <= dbot):
```
andy-beer, line 599: This will include cells that are fully within the dtop to dbot depth range. It would be better if we included cells that are partially in the range, though ideally with a PPERF being set too. Also, it might be slightly better to use the K-, K+ face mid points, rather than the corner points, as the definition of the cell depth range for this purpose.

The corp z units might be increasing upwards (ie. negative values), in which case the z values should be negated before applying this test. You can check the z axis direction in the boolean attribute: grid.crs.z_inc_down which is usually True but not always.

andy-beer:
Thanks for picking up on this missing functionality Patrick, and for working up the fix. A few comments responding to your questions and other thoughts that came to me:

- The Grid.corner_points() method does have an optional cache_cp_array argument which defaults to False but can be used to cache a copy of the corner point array in the Grid object, if appropriate.
- Creating the full corner point array can be a bit slow and very memory intensive. For this WELLSPEC functionality, it might be better to call it repeatedly for one cell at a time, by passing the optional cell_kji0 argument, which causes it to only do the computations for individual cells rather than the whole array.
- You might consider using the Grid.face_centre() method instead of corner_points(), with the K face centres being used to define the depth range of a cell.
- If we are not setting PPERF, and having an all-or-nothing approach to each cell, then an alternative would be to use the Grid.centre() method and decide a cell's inclusion based solely on the depth of the centre point.
- We have a very stringent unit test code coverage regime these days – any changes need to be accompanied with unit tests to at least match the previous average coverage (currently around 81%).


WELLSPEC is a Nexus keyword. Nexus is a trademark of Halliburton.

PR #788, opened 2024-02-24
Symmetric hexagonal stencil class for convoluting tri meshes

andy-beer:
This change adds a TriMeshStencil class which can be used to smooth (or apply other convolutions) to TriMesh z values.


andy-beer:
unit tests not yet added

PR #797, opened 2024-04-17
FIX: allows yearOffset at 0 for geologic time series

adamchengtkc:
Currently Geological time series can't be 0 as this line check for `falsy`. Suggesting this PR to check for `int` instead.
User case: Storing values that goes from geological past to present-day which is denoted as `0`


andy-beer:
Thanks for this change. I am going to replicate it in a separate branch as the CodeCov aspect of the unit test pipeline is currently falling over when running from the equinor:master branch. (I will also ask a colleague to look into whether we can get that working, for future changes.)

andy-beer:
This Pull Request has been replaced by #798 which is functionally equivalent and has been merged into master. I am therefore closing this PR without merging. FAO. @adamchengtkc

PR #856, opened 2024-11-08
New surface method: Resample surface edges/tears

emmanesbit:
New method to refine triangles on the edges/tears within triangulated sets (resqpy.surface.Surfaces).


andy-beer (approved):
This is a neat piece of coding.

At a later date we could consider adding an option to avoid retriangulation, ie. preserving tears. For that we would need to replace the edge triangle point indices with data for a bunch of explicitly created replacement triangles.

Another nice-to-have would be the ability to specify how many extra points to add per original tear edge – especially as retriangulation means that this function cannot be called multiple times to add more points along a tear.

Finally, it would be good to have an option to differentiate tears (including open-ended tears) from the outer hull of the surface, and then only add extra points for the tears.

PR #862, opened 2024-12-06
New method to create and flange extended surface, avoiding retriangulation

emmanesbit:
New method allowing a surface to be generated based on an existing surface, with flange extended points. The new triangles and points are added directly, avoiding retriangulation.


resqpy/surface/_surface.py
```diff
@@ -706,6 +707,187 @@ def set_from_point_set(self,
...
+            )  # loop over all the hull centres, generating a flange point and finding the azimuth from the centre to the hull centre point
+            az = np.empty(shape = len(hull_centres), dtype = float)
+            for i, c in enumerate(hull_centres):
+                v = [centre_point[0] - c[0], centre_point[1] - c[1], centre_point[2] - c[2]]
+                uv = -vec.unit_vector(v)
+                az[i] = vec.azimuth(uv)
+                flange_point = centre_point + radius * uv
+                if simple_saucer_angle is not None:
+                    z_shift = radius * maths.tan(vec.radians_from_degrees(simple_saucer_angle))
+                    flange_point[2] -= z_shift
```
andy-beer, line 825: if we have not reoriented, then we should apply this 'z_shift' in the direction of the normal vector for the surface, rather than along the z axis.  In other words, get the normal vector (as a unit vector) and multiply it by the z_shift value, to give the xyz shift to subract from the flange points. (Add or subtract depending on the direction of the normal vector, as indicated by the sign of its z component.)

PR #865, opened 2024-12-18
Allow 0 YearOffset elements for 'human' timescales

Gary-Lockyer:
Allow the YearOffset to be zero in TimeSeries with a "human" timescale. Have encountered examples of this, and it seems a reasonable thing to allow.


andy-beer:
This change looks good to me, thanks. The code coverage testing failed on a data rate limit set by the codecov service – not related to the change here. I'll get the change merged once codecov is happy again.

PR #867, opened 2024-12-18
Remove assert in extract_stratigraphy

Gary-Lockyer:
replace the assertion:
  len(grid.stratigraphic_units) == grid.nk_plus_k_gaps
with a warning message

Have encountered example files that fail this assertion.

This allowed the processing of the affected files, but I'm not sure what other impacts this change could have.

### DIFF
--- a/resqpy/grid/_extract_functions.py
+++ b/resqpy/grid/_extract_functions.py
@@ -445,7 +445,12 @@ def extract_stratigraphy(grid):
                                 object = grid,
                                 array_attribute = 'stratigraphic_units',
                                 dtype = 'int')
-    assert len(grid.stratigraphic_units) == grid.nk_plus_k_gaps
+    if (len(grid.stratigraphic_units) != grid.nk_plus_k_gaps):
+        log.warning(
+            "Number of stratigraphic units (%d) does not equal the number of layers plus gaps (%d)",
+            len(grid.stratigraphic_units),
+            grid.nk_plus_k_gaps,
+        )
 
 
 def extract_children(grid):


andy-beer:
Hmm, I checked the RESQML schema, and Grids.xsd includes the following documentation for IntervalStratigraphicUnits:

_Index of the stratigraphic unit per interval, of a given stratigraphic column.

Notes:
1.) For grids, intervals = layers + K gaps.
2.) If there is no stratigraphic column, e.g., within salt, use null (-1)

BUSINESS RULE: Array length must equal the number of INTERVALS._

So it looks like the files you encountered (@Gary-Lockyer) were not in accordance with this business rule. I will check the resqpy code but I suspect that it requires this mapping to intervals if the stratigraphy is to be associated with grid layers.

andy-beer:
A follow up: As I suspected, the number of stratigraphic unit entries must match the number of intervals in the grid for the data to be usable – the whole point is to establish the mapping from grid layers to stratigraphy. So, I propose that we don't use the change as you have prepared it @Gary-Lockyer. Instead, if the len(grid.stratigraphic_units) != grid.nk_plus_k_gaps, let's log an **error** level message (instead of warning), also reset the two Grid attributes to None (as in the first two lines of the function). That way the Grid will load as if the stratigraphy information is not there (it is optional data anyway). I think that is likely to cause fewer problems than having an array of incorrect size in the resqpy data structures. Any thoughts?

I'd also encourage you to mention this issue to whomever supplied the data, so that their software can be improved.

resqpy/grid/_extract_functions.py
```diff
@@ -445,7 +445,12 @@ def extract_stratigraphy(grid):
                                 object = grid,
                                 array_attribute = 'stratigraphic_units',
                                 dtype = 'int')
-    assert len(grid.stratigraphic_units) == grid.nk_plus_k_gaps
+    if (len(grid.stratigraphic_units) != grid.nk_plus_k_gaps):
+        log.warning(
+            "Number of stratigraphic units (%d) does not equal the number of layers plus gaps (%d)",
```
andy-beer, line 450: We have been using the f"blah blah {expression} blah {expression}" form of string formatting elsewhere in resqpy. For consistency it might be nice to stick with that convention, though it is not particularly important, if you feel strongly about the % way of formatting!

andy-beer:
Thanks for making these changes, which look good to me.

Unfortunately the unit test does not seem to be covering the if clause as intended. Perhaps the s_bend dataset does not include interval stratigraphic data? (I haven't checked.)

Gary-Lockyer:
s_bend -does not have any stratigraphy, I moved the test to strata/test_strata.py, to avoid block copying the grid set-up code.  However I'm not convinced that's the correct place.

andy-beer (approved):

andy-beer:
Thanks for making those changes Gary. I've merged this pull request and will be making a new resqpy release later today.

PR #869, opened 2024-12-19
MAINT: Fix codecov upload, bump github action tags

connortann:
Fix the codecov action by bumping the version, and updating the repo upload token.

Reset `fail_ci_if_error: true`.

Also bump outdated `upload-artifact`, `cache` and `setup-python` actions.


andy-beer (approved):
This looks great Connor, thanks! I had also changed a yes to no in .github/codecov.yml – should we change that back to yes at the same time?
codecov:
  require_ci_to_pass: no

PR #870, opened 2024-12-20
Updates to ensure reorientation before flange point generation, even where triangulation is completed in non-reoriented space

emmanesbit:
These changes revert previous changes using the non-triangulation method when a saucer is required, but reorient is False. Instead the set_from_point_set method can now handle saucers without reorientation.

These changes mean that when a flange points are being added, but reorientation is False, the flange points are still generated and added based on the reoriented points (to the normal vector orientation), however the triangulation is completed in non-reoriented space

### DIFF
--- a/resqpy/surface/_surface.py
+++ b/resqpy/surface/_surface.py
@@ -636,7 +636,7 @@ class Surface(rqsb.BaseSurface):
             assert -90.0 < saucer_parameter < 90.0, f'simple saucer angle parameter must be less than 90 degrees; too big: {saucer_parameter}'
             simple_saucer_angle = saucer_parameter
             saucer_parameter = None
-        assert saucer_parameter is None or 0.0 <= saucer_parameter < 1.0
+        assert saucer_parameter is None, 'saucer_parameter no longer supported, use simple_saucer_angle'
         crs = rqc.Crs(self.model, uuid = point_set.crs_uuid)
         p = point_set.full_array_ref()
         assert p.ndim >= 2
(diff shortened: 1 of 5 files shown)


resqpy/surface/_surface.py
```diff
@@ -654,32 +654,23 @@ def set_from_point_set(self,
...
-                                                            radial_distance = flange_radial_distance,
-                                                            inner_ring = flange_inner_ring,
-                                                            saucer_angle = simple_saucer_angle)
-            p_xy_e = np.concatenate((p_xy, flange_points), axis = 0)
+            flange_points, radius = triangulate.surrounding_xy_ring(p_xy,
+                                                                    count = flange_point_count,
+                                                                    radial_factor = flange_radial_factor,
+                                                                    radial_distance = flange_radial_distance,
+                                                                    inner_ring = flange_inner_ring,
+                                                                    saucer_angle = 0)
```
andy-beer, line 665: could you make it saucer_angle = 0.0 please – I'm still fussy about variable types even if python is not

resqpy/surface/_surface.py
```diff
@@ -694,15 +685,20 @@ def set_from_point_set(self,
             t = triangulate.dt(p_xy_e[:, :2], container_size_factor = convexity_parameter * 1.1)
         log.debug('number of triangles: ' + str(len(t)))
         if make_clockwise:
```
andy-beer, line 687: Let's make make_clockwise True in a unit test, so the make_all_clockwise_xy() gets to run at least. It shouldn't change any behaviour of the code afterwards.

resqpy/surface/_surface.py
```diff
@@ -636,7 +636,7 @@ def set_from_point_set(self,
             assert -90.0 < saucer_parameter < 90.0, f'simple saucer angle parameter must be less than 90 degrees; too big: {saucer_parameter}'
             simple_saucer_angle = saucer_parameter
             saucer_parameter = None
-        assert saucer_parameter is None or 0.0 <= saucer_parameter < 1.0
+        assert saucer_parameter is None, 'saucer_parameter no longer supported, use simple_saucer_angle'
```
andy-beer, line 639: My mistake here, the argument is still saucer_parameter. We can remove this assertion completely.

resqpy/surface/_surface.py
```diff
@@ -636,7 +636,7 @@ def set_from_point_set(self,
             assert -90.0 < saucer_parameter < 90.0, f'simple saucer angle parameter must be less than 90 degrees; too big: {saucer_parameter}'
```
andy-beer, line 636: Let's strip out the old option, so remove the (saucer_parameter > 1.0 or saucer_parameter < 0.0) please. Also, please update the docstring notes to strip out the old option (1)

resqpy/surface/_surface.py
```diff
@@ -694,15 +685,20 @@ def set_from_point_set(self,
...
             flange_array[:] = np.where(np.any(t >= len(p), axis = 1), True, False)
-            if saucer_parameter is not None:
-                _adjust_flange_z(self.model, self.crs_uuid, p_xy_e, len(p), t, flange_array, saucer_parameter)
-                p_e = vec.rotate_array(reorient_matrix.T, p_xy_e)
-        if crs.xy_units != crs.z_units and reorient:
-            wam.convert_lengths(p_e[:, 2], crs.xy_units, crs.z_units)
+            if simple_saucer_angle is not None:
+                assert abs(simple_saucer_angle) < 90.0
+                z_shift = radius * maths.tan(vec.radians_from_degrees(simple_saucer_angle))
+                flange_points[:, 2] -= z_shift
```
andy-beer, line 689: Yes, this -= is consistent with how the argument is described.

resqpy/surface/_surface.py
```diff
@@ -694,15 +685,20 @@ def set_from_point_set(self,
...
         if make_clockwise:
-            triangulate.make_all_clockwise_xy(t, p_e)  # modifies t in situ
+            triangulate.make_all_clockwise_xy(t, p_xy_e)  # modifies t in situ
         if extend_with_flange:
             flange_array = np.zeros(len(t), dtype = bool)
             flange_array[:] = np.where(np.any(t >= len(p), axis = 1), True, False)
-            if saucer_parameter is not None:
-                _adjust_flange_z(self.model, self.crs_uuid, p_xy_e, len(p), t, flange_array, saucer_parameter)
-                p_e = vec.rotate_array(reorient_matrix.T, p_xy_e)
-        if crs.xy_units != crs.z_units and reorient:
```
andy-beer, line 704: And this now needs to be just:
if crs.xy_units != crs.z_units:
I think!

andy-beer (approved):
Please address the various comments I've added, then good to merge if you are happy with it.

PR #888, opened 2025-04-15
Support Python 3.13

cflynn3:


connortann:
Support for 3.13 added on !905

PR #902, opened 2026-05-20
Update _grid_connection_set.py for stable sorting

emmanesbit:
Previously tests were failing on occasion as the sorting was not stable and so different results were possible where a cell face could be claimed by two triangles. Stable sorting aims to avoid this ambiguity


cflynn3 (approved):
LGTM. Hopefully this fixes the issue.

PR #904, opened 2026-06-11
Drop support for python 3.9, update deprecated github action versions & poetry version

connortann:
Several github actions are deprecated and are no longer supported. This MR updates all actions and the poetry version to something recent. In turn this requires dropping support for python 3.9 , which is past EOL and no longer supported by recent poetry.

Updates a fair few deprecated dependencies:
- Updates all github actions to latest version; many current ones use deprecated node runtimes and are not supported
- Update the installed poetry version to a recent version
- Drops support for python 3.9

Unblocks a follow-up MR that adds support for 3.13 and 3.14 : #905


emmanesbit (approved):
Looks good to me

PR #907, opened 2026-06-12
Add pre-commit config

### LINKED ISSUE #394: Enable pre-commit hooks
Plenty of PRs have failed jobs due to code formatting and flake8 errors, which suggests that some time is being wasted by devs having to run these tools manually.

We could use the `pre-commit` python package as a developer dependency, and add a `.pre-commit-config.yaml` file together with some instructions in the contributor guide to install.

This would mean yapf is run locally automatically on relevant files before each commit is made, so devs can spend their time on more important things.

connortann:
Adds a `pre-commit` config, to orchestrate static analysis checks (linting & formatting).

Closes #394

Some benefits:
- It runs on CI automatically via pre-commit.ci , so we don't need to maintain a separate GH actions job
- Allows devs to run linting and formatting checks locally with identical versions to what is run on CI
- We get other simple checks for free (e.g. prevents mixed line endings, or merge commit markers)

### DIFF
--- a/.github/workflows/ci-tests.yml
+++ b/.github/workflows/ci-tests.yml
@@ -9,8 +9,8 @@ on:
   pull_request:
 
 jobs:
-  static-analysis:
-    name: Static analysis
+  type-check:
+    name: Type check
     runs-on: ubuntu-latest
     steps:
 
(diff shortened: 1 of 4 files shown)


emmanesbit (approved):
LGTM

PR #908, opened 2026-06-12
Update linter config to ruff

connortann:
### Summary

A small housekeeping PR.

- Updates our linting config to use the modern "ruff" implementation, retaining broadly the same set of linting rules as before.
- A few minor whitespace and variable name fixups to appease the latest version of these checks: in particular, replacing the use of `l` as a single-character variable name in a few places.

Chained upon #394 which should be merged first.

### Discussion

Motivation for updating to ruff:

- 10-100x faster
- Supports configuration in the standard pyproject.toml
- Consistent with most modern python projects


emmanesbit (approved):
LGTM

PR #909, opened 2026-06-15
Remove unused imports (auto-fixed by ruff), reducing circular imports

connortann:
A rather trivial edit, to help reduce occurrence of circular imports.

- Removes all unused imports
- Enables the linting rule "F401"

Although the diff is large, it was generated almost entirely mechanically with ruff, so it's simple to re-create.

```bash
ruff check --select=F401 --fix
```

The only case that requires manual attention is when an object is imported into a module, with the intention that is available there to be re-imported somewhere else. In particular:
- `xml_et.py` : exposes variables from `lxml.etree`
- `__init__.py` files: import objects from private modules to be re-exported. Fix is to add them to the `__all__` list of public variables.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -111,7 +111,6 @@ select = [
     "D",                         # pydocstyle (Google-style docstrings)
 ]
 ignore = [
-    "F401",  # module imported but unused: low-priority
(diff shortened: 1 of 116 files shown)


emmanesbit (approved):
LGTM

PR #912, opened 2026-06-15
Allow configuration of numba cache via environment variable

connortann:
If `RESQPY_NUMBA_CACHE=TRUE`, enable numba caching. If unset, the cache is not enabled, which is numba's default behaviour if the `cache` keyword is not passed.

Caching leads to a significant speedup on first use on startup - it can lead to around ~30% speedup in other projects' test suites. However, we probably will wish to disable the cache when developing njit-decorated functions, to ensure that we don't use stale compiled artifacts.

Having this available as an environment variable makes it easier to experiment downstream. The default behaviour of resqpy is unchanged.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -138,7 +138,8 @@ module = [
     "numpy",
     "lasio",
     "h5py",
-    "lxml"
+    "lxml",
+    "numba"
 ]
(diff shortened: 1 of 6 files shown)


emmanesbit (approved):
LGTM

PR #913, opened 2026-06-15
Add 3rd party type stubs to dev dependencies

connortann:
## Summary

Adds some 3rd party stub files to dev dependencies, for more accurate type checking.

## Discussion

Installing these typing stubs is an easy way to improve the accuracy of mypy.

As a side-note: at the moment mypy ignores the majority of resqpy, as untyped functions are ignored. But given that we run mypy, we may as well provide complete type information.

One subtlety is that pandas-stubs versions are not directly tied to pandas versions - it is the [recommendation of the stubs authors ](https://github.com/pandas-dev/pandas-stubs/issues/1724) to use the latest version of pandas-stubs, regardless of the installed version of pandas in use.


cflynn3 (approved):
Looks like a good idea. A good first issue for someone would be to go through the code and add all the missing types everywhere.

PR #914, opened 2026-06-16
MAINT: Fixup flaky test in grid surface tests and fixup warnings

connortann:
One test was failing intermittently locally, due to slightly different random peturbations:

```bash
tests/unit_tests/test_grid_surface.py::test_find_faces_to_represent_surface_regular_optimised_random_agitation
```

Fixes this test by allowing small deviations due to peturbations, and also supresses some expected internal Deprecation warnings.

As of this PR, the resqpy tests suite pasees on python 3.14 with zero warnings.

### DIFF
--- a/tests/unit_tests/test_grid_surface.py
+++ b/tests/unit_tests/test_grid_surface.py
@@ -319,7 +319,6 @@ def test_find_faces_to_represent_surface_regular_optimised_random_agitation(smal
                                                               agitate = True,
                                                               random_agitation = True)
     cip_normal = gcs_normal.cell_index_pairs
-    fip_normal = gcs_normal.face_index_pairs
 
     gcs_optimised = rqgs.find_faces_to_represent_surface_regular_optimised(grid,
                                                                            surface,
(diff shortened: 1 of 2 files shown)


tests/unit_tests/test_grid_surface.py
```diff
@@ -312,19 +319,22 @@ def test_find_faces_to_represent_surface_regular_optimised_random_agitation(smal
                                                               agitate = True,
                                                               random_agitation = True)
     cip_normal = gcs_normal.cell_index_pairs
-    fip_normal = gcs_normal.face_index_pairs
```
cflynn3, line 315: Please can you explain why we are no longer checking the face index pairs?

tests/unit_tests/test_grid_surface.py
```diff
@@ -312,19 +319,22 @@ def test_find_faces_to_represent_surface_regular_optimised_random_agitation(smal
...
                                                                            random_agitation = True)
     cip_optimised = gcs_optimised.cell_index_pairs
-    fip_optimised = gcs_optimised.face_index_pairs
 
-    # Assert
-    np.testing.assert_array_equal(cip_normal, cip_optimised)
-    np.testing.assert_array_equal(fip_normal, fip_optimised)
+    # The two functions agitate the surface with random peturbations so may not match exactly.
+    # Assert near-agreement by comparing cell index pairs but allowing a small symmetric difference.
+    pairs_normal = {tuple(pair) for pair in cip_normal}
```
cflynn3, line 332: I think this is a good way of handling the current flakiness, but is there not a way we can limit the 'randomness' by seeding the random function, perhaps?

cflynn3 (approved):
LGTM.

tests/unit_tests/test_grid_surface.py
connortann: We can use `np.random.seed` to seed the local state; however, in general the numpy random generator will behave differently between python versions, numpy versions and architectures, so I don't think it will make any difference to test flakiness.

FWIW numpy has some recommendations on handling randomness, which involve accepting a `rng` argument to relevant functions to control seeding, rather than using the global numpy random state. I'm not sure anyone needs that functionality, but it should be easy enough to implement if there is demand.
https://github.com/numpy/numpy/blob/main/doc/neps/nep-0019-rng-policy.rst

PR #915, opened 2026-06-17
Add yapf:disable comments for data blocks

connortann:
A tiny housekeeping change: adds a few yapf:disable and yapf:enable comments around data blocks. No change for the current code, but if we run `ruff format` in future, these blocks will not be changed.

Proposing as a separate PR just to keep any future "ruff format" MR a simple one-command change.


emmanesbit (approved):
LGTM

PR #916, opened 2026-06-18
Add dependabot config

connortann:
Add dependabot config to scan for new dependencies and open MRs with and updates lockfile.

Reference: https://docs.github.com/en/code-security/how-tos/secure-your-supply-chain/secure-your-dependencies/configure-version-updates

We'll probably want this running "monthly" or "quarterly" frequency, but proposing running it "daily" to start with so we can test and verify the behaviour.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,30 @@
+---
+
+version: 2
+updates:
+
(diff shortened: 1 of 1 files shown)


emmanesbit (approved):
Looks good to me

PR #925, opened 2026-06-22
Loosen lxml lower bound

connortann:
Resqpy still works with lxml v5, so in the spirit of SPEC0 we can keep that as an allowed version. There do not seem to have been significant breaking changes in lxml v6, other than dropping support for python 3.8.


cflynn3 (approved):
LGTM.

PR #926, opened 2026-06-22
MAINT: Better cleanup of temporary test files

connortann:
Use the `tmp_path` fixture to ensure temporary wrapper.epc files are created and cleaned up in the standard clean manner, rather than appearing in the current working directory

Supporting minor change: the `poetry lock` check is occasionally failing when running on pre-commit.ci, due to a flaky network connection with PyPI. So, moving that check to run under GH actions instead, as part of the poetry env install.


cflynn3 (approved):
Looks good to me.

